PR #16, opened 2022-06-15
Turkish Localization

Hexasan:
Added Turkish Language File
Hexasan#0152


src/main/resources/assets/emi/lang/tr_tr.json
```diff
@@ -0,0 +1,411 @@
...
+    "config.emi.general.enabled": "Aktif",
+    "config.emi.general.cheat_mode": "Hile Modu",
+  
+    "config.emi.ui.move_effects": "Hareket Efektleri",
+    "config.emi.ui.center_search_bar": "Arama Kutucuğunu Ortala",
+    "config.emi.ui.show_hover_overlay": "Üstünden Geçirme Kaplamasını Göster",
+    "config.emi.ui.append_item_mod_id": "Eşya Mod ID'sini Ekle",
+    "config.emi.ui.empty_search_craftable": "Boş aramayla Üretilebilenleri Göster",
+    "config.emi.ui.vertical_margin": "Dikey  Sınırı",
+    "config.emi.ui.miscraft_prevention": "Miscraft Engellemesi",
```
unascribed, line 41: I notice "Miscraft" is kept without being translated; "miscraft" refers to crafting an item by mistake. This can be kept but I wanted to call it out in case it's only not translated due to not understanding the source word, since it's technically made up.

A more translatable source string would be "Accidental Crafting Prevention".

unascribed (approved):
Once-over with a machine translator looks OK (due diligence, checking for profanity/etc) - one question, though

unascribed (approved):

PR #22, opened 2022-06-20
Hopeful fix for #21

### LINKED ISSUE #21: Crash if maxDamage = 0
Sorry if this is just me being a terrible programmer but when making a tool/armour unbreakable I tend to give it a max damage of 0, but when EMI tries to render the anvil screen that causes issues with the random.nextInt() since the bound must be positive.

[crash-2022-06-20_04.57.11-client.txt](https://github.com/emilyploszaj/emi/files/8937407/crash-2022-06-20_04.57.11-client.txt)

AmyMialee:
Should hopefully fix #21.

### DIFF
--- a/src/main/java/dev/emi/emi/VanillaPlugin.java
+++ b/src/main/java/dev/emi/emi/VanillaPlugin.java
@@ -281,10 +281,12 @@ public class VanillaPlugin implements EmiPlugin {
 		}
 
 		for (Item i : Registry.ITEM) {
-			if (i instanceof ArmorItem ai) {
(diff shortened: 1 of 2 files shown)


emilyploszaj:
Sorry to bikeshed such a small PR, but shouldn't a tool that has a max damage of zero just not get an anvil recipe?

AmyMialee:
Yea you're definitely right, did it after getting some sleep.
I kept the old one as a failsafe, just in case some mod does some wacky mixins like making the Elytra have 0 maxDamage or other similar issues.

PR #33, opened 2022-07-13
Chinese simplified translation

IJustWantANamePlease:
zh_cn


emilyploszaj:
Sorry for the late reply, yes, all of the config tooltips are translatable. They use the hardcoded strings from the config file by default, but can use `config.emi.tooltip` at the start instead `config.emi`, for example, `config.emi.tooltip.ui.move_effects`

IJustWantANamePlease:
> Sorry for the late reply, yes, all of the config tooltips are translatable. They use the hardcoded strings from the config file by default, but can use `config.emi.tooltip` at the start instead `config.emi`, for example, `config.emi.tooltip.ui.move_effects`

Thanks, I get all the config working. However, this doesn't seem to apply to hotkeys. I tried adding "tooltip." but it doesn't work. Is there a difference in hotkey tooltips?

emilyploszaj:
Hotkeys use a `key.emi.view_recipes` format or similar, but that is in the normal en_us

IJustWantANamePlease:
> Hotkeys use a `key.emi.view_recipes` format or similar, but that is in the normal en_us

I mean tooltips for hotkeys, I tried to change them using the same method but failed.

IJustWantANamePlease:
Oh wait, I get it working, sorry for bothering.

IJustWantANamePlease:
"tooltip.emi.resolution": "Resolving recipe tree for ingredient",
May I as what this means? I can't find this tooltip in the game to get the context.
Also, this might be a bug, the process name for tag isn't displayed in receipe tree:
图片

emilyploszaj:
tooltip.emi.resolution is used when you've clicked on an item in the recipe tree and are browsing recipes, it should be in the upper left with some colors. Also, resolution isn't supposed to show up like that in game, how did you make that happen?

IJustWantANamePlease:
> tooltip.emi.resolution is used when you've clicked on an item in the recipe tree and are browsing recipes, it should be in the upper left with some colors. Also, resolution isn't supposed to show up like that in game, how did you make that happen?

This happens when I right click on this node after resolving an receipe.
图片

emilyploszaj:
I was doing a pass for profanity and was going to merge, but I ran into something peculiar, all of the translations are unicode escaped. Does the minecraft json format have issues with simplified chinese characters? Or is this standard for Minecraft translations. Either way, it looks great to me, and I'm going to merge after we clear this up.

IJustWantANamePlease:
Well, I believe there is no issue Minecraft dealing with Chinese characters, I haven't encountered any at least. The reason of me using unicode is that the translation of Minecraft itself still does this (and I just blindly follow that). It's probably not required, many mod and resource packs uses Chinese characters directly. I could upload the "Chinese" version if you want.

unascribed:
Minecraft always loads language JSONs with the UTF-8 charset, so from that perspective it's perfectly safe. People on Windows with busted system configs may accidentally corrupt these files simply by breathing on them, and a lot of people use Windows, so that's generally the argument in favor of escaping. It's also fairly likely that whatever tool Mojang uses to export the translations from Crowdin automatically escapes things.

For ease of editing, I'd suggest leaving it unescaped. PRs that corrupt the lang file can be rejected should they show up, but Windows too has been getting better about not trouncing UTF-8 files.

IJustWantANamePlease:
Sure, I've switched it.

emilyploszaj:
Hopefully readable characters means this is easier to update in the future. Thank you so much for this contribution, it seems to be a really excellent translation. I hope this makes EMI more accessible for new users.

PR #52, opened 2022-07-31
Added missing Firework recipes

jopejoe1:
This adds the missing firework recipes I mentioned in #48
Closes #48

PR #53, opened 2022-07-31
Added missing recipes and recipe fixes

jopejoe1:
This adds the following recipes:

- Banner Duplication
- Map Cloning
- Item Repair (Crafting Table version)

This fixes:

- Map extending having empty map as recipe result
- Re adds one line I miss deleted in https://github.com/emilyploszaj/emi/pull/52/commits/fab9d6d065ce2eb550cca8182f625762542635ee
- Banner Shields not being listed as a  recipe when clicking on shields


emilyploszaj (approved):
Another great PR, thank you so much for your contributions. A note for the future, recipes that do not support the recipe tree do not have to be super accurate with their inputs and outputs (I saw you took caution in this PR and the previous to set accurate ingredients and remainders), if they can't be filled or expressed in a recipe tree, inputs and outputs are only context for recipe lookups, but I do appreciate being thorough, so I'll keep it haha.

PR #61, opened 2022-08-14
Added more recipes to anvils

jopejoe1:
This adds repair with tool and enchanting with enchanting book to the recipe list of anvils.

### DIFF
--- a/src/main/java/dev/emi/emi/VanillaPlugin.java
+++ b/src/main/java/dev/emi/emi/VanillaPlugin.java
@@ -48,6 +48,8 @@ import dev.emi.emi.recipe.EmiShapelessRecipe;
 import dev.emi.emi.recipe.EmiSmithingRecipe;
 import dev.emi.emi.recipe.EmiStonecuttingRecipe;
 import dev.emi.emi.recipe.EmiTagRecipe;
+import dev.emi.emi.recipe.special.EmiAnvilEnchantRecipe;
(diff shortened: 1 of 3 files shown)


emilyploszaj (commented):
Hm. I'm concerned about the bloat and perf implications of this PR. What kind of reload time does this incur in vanilla (and will it scale unreasonably to larger modpacks)? Also, curse of vanishing can be applied to every item in the game, am I right in thinking this will mean there will be a recipe for every item adding curse of vanishing?

jopejoe1:
> I'm concerned about the bloat and perf implications of this PR.

Didn't really think about that as i was making this PR, did some testing and showing what i found below. Hope that helps.


> What kind of reload time does this incur in vanilla

In vanilla it takes about ~40 ms to reload compared to ~30 ms before. 


> and will it scale unreasonably to larger modpacks

I also tested this with around a 100 mods where the reload time increased from ~400 ms to ~500 ms.


> Also, curse of vanishing can be applied to every item in the game, am I right in thinking this will mean there will be a recipe for every item adding curse of vanishing?

No, curse of vanishing can only be applied to equipment and compass, there was once a snapshot where it applied to everything but that was fixed in the full release.

emilyploszaj:
Oh, interesting, good to know. a +33% perf hit is pretty unfortunate but it's not unreasonable I suppose. Thanks for the PR.

PR #67, opened 2022-08-17
Added tall flower dupe recipe 

jopejoe1:
Added the tall flower dupe recipe to the list of world interaction recipes.


emilyploszaj (approved):
Nice catch! Thanks again for your contribution

PR #71, opened 2022-08-20
Chinese Simplified Update.

IJustWantANamePlease:
This is the updated translation for 0.3.4. 
I wasn't going to PR since there aren't much added stuff, until I notice there are two lines which require correction.

Also, are there any chance that EMI will be using CrowdinTranslate for translation management? In this way, older versions can still receive translation update after they have been published.


emilyploszaj (approved):
Unfortunately, in regards to Crowdin, I'm not sure I have the development time to organize on a separate platform, though, it will be a consideration for the future. At the moment I am attempting to release regular updates with fixes and translations, and as EMI stabilizes this should hopefully become less pertinent. Either way, I appreciate your work further translating EMI.

PR #72, opened 2022-08-21
Added Grindstone recipes

jopejoe1:
Added Grindstone recipe category and recipes.


emilyploszaj (approved):
Looks good to me. Couple questions, looks like you reused the simplified texture for an anvil, I assume that's a placeholder for me to resolve, which is probably reasonable, is that it?. Also, I was expecting a generated recipe for grindstones. The upsides to this solution is it provides xp return breakdown, but the downside is it doesn't show explicitly that curses get retained. I think this is fine, but I'd just like your thoughts.

jopejoe1:
> looks like you reused the simplified texture for an anvil, I assume that's a placeholder for me to resolve, which is probably reasonable, is that it?

Ops, I forgot that,

> Also, I was expecting a generated recipe for grindstones. The upsides to this solution is it provides xp return breakdown, but the downside is it doesn't show explicitly that curses get retained.

Maybe could make all the tools into generated recipes and keep the enchanted books with only one enchantment to show how much xp is returned.

emilyploszaj:
Oh, that split tool/book solution is a good idea, I think that'd be the most intuitive.

jopejoe1:
Added the split between enchanted book and tools.

emilyploszaj (approved):

PR #74, opened 2022-08-23
Tall flower dupe fix

### LINKED ISSUE #73: Tall flower world interaction recipe wrongly matches botania tall flowers
image
prank recipe not real
cc @jopejoe1

jopejoe1:
Fixes some modded Tall Flowers being matched even tough they can not be duplicated.

Fixes #73

### DIFF
--- a/platform-1.18.2/src/main/java/dev/emi/emi/EmiPort.java
+++ b/platform-1.18.2/src/main/java/dev/emi/emi/EmiPort.java
@@ -13,6 +13,7 @@ import com.mojang.brigadier.CommandDispatcher;
 import net.fabricmc.fabric.api.command.v1.CommandRegistrationCallback;
 import net.fabricmc.fabric.api.transfer.v1.client.fluid.FluidVariantRendering;
 import net.fabricmc.fabric.api.transfer.v1.fluid.FluidVariant;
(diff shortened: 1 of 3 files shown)


emilyploszaj (changes requested):
None of those params are nullable as far as I can tell, passing the default state in a try catch block would probably be safer. Though, I think these method signatures break on the 1.18/1.19 boundary and will crash on a version update because of the changed random instances, we might just want to hardcode this for now.

jopejoe1:
> None of those params are nullable as far as I can tell, passing the default state in a try catch block would probably be safer.

Moved them into a catch block.

> I think these method signatures break on the 1.18/1.19 boundary and will crash on a version update because of the changed random instances,

Yeah didn't really test this on 1.19 as botania wasn't available on it and i was using it as a filter test.

> we might just want to hardcode this for now.

Moved the logic into emiport as i don't really want to make a pr to all the mods that also add tall flowers to also add them to the list. (like byo and biome makeover and others i don't even know)

> I'd do it myself but you already have a PR open :p

:p

jopejoe1:
If you still would prefer it to be hardcoded than having it in emiport i will switch it over.

emilyploszaj:
Nah this is fine, I'll tinker with it myself if it causes issues

PR #97, opened 2022-11-21
Fix window resizing with Fabrications old_tooltip

SFort:
Fixes /unascribed/fabrication#573

### DIFF
--- a/src/main/java/dev/emi/emi/mixin/ScreenMixin.java
+++ b/src/main/java/dev/emi/emi/mixin/ScreenMixin.java
@@ -1,21 +1,38 @@
 package dev.emi.emi.mixin;
 
-import java.util.List;
-
-import org.spongepowered.asm.mixin.Mixin;
-import org.spongepowered.asm.mixin.Shadow;
-import org.spongepowered.asm.mixin.Unique;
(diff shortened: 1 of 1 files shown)


emilyploszaj:
This feels liable to break mods that wrap tooltips and try to use introspection to place their tooltips, like botania's mana. Spoofing the width is the only reliable way to deal with these cases as far as I can tell.

SFort:
i'm assuming this renders correctly i haven't played modded in a while. so you'll have to tell me a broken example


https://user-images.githubusercontent.com/1879846/202981832-c3099a8e-f103-4c98-b328-afc60f7851bd.mp4

emilyploszaj:
The specific error occurs when the tooltip has space to move to the right, but EMI is forcing it to the left, this is easier to see on lower GUI scales. Overall though, I think this is just continuing down a road of hacking a problem, when I could easily add a helper method that does the following.
```java
int original = screen.width;
screen.width = mouseX;
try {
  renderTooltip();
} finally {
  screen.width = original;
}
```

emilyploszaj:
My current local changes have a helper method for rendering tooltips, if this approach is agreeable I can close this PR and fix it locally for the next release.

PR #99, opened 2022-11-22
Recipe Screenshots

Kneelawk:
# This PR
This PR adds the ability to take screenshots of recipes from EMI.

First, the screenshot button must be enabled from the EMI config file. Then, screenshots can be taken of recipes by clicking the screenshot button to the left of the recipe. The scale of screenshots is, by default, the current UI scale of the user's game, but can be set to a specific value in the EMI config file.

## Screenshots
recipe-2022-11-22_07 00 00


src/main/java/dev/emi/emi/screen/RecipeScreen.java
```diff
@@ -358,6 +355,9 @@ public <T extends Widget> T add(T widget) {
 						widgets.add(new RecipeDefaultButtonWidget(r.getDisplayWidth() + 5, by - 14 * button++, r));
 					}
 				}
+				if (EmiConfig.recipeScreenshotButton) {
+					widgets.add(new RecipeScreenshotButtonWidget(-5 - 12, 0, r));
```
emilyploszaj, line 359: Screenshot button should be aligned to the bottom of the recipe like the buttons on the right, what do you think?

src/main/java/dev/emi/emi/widget/RecipeScreenshotButtonWidget.java
```diff
@@ -0,0 +1,33 @@
...
+		return List.of(TooltipComponent.of(EmiPort.ordered(EmiPort.translatable("tooltip.emi.recipe_screenshot"))));
+	}
+
+	@Override
+	public boolean mouseClicked(int mouseX, int mouseY, int button) {
+		this.playButtonSound();
+
+		int width = recipe.getDisplayWidth() + 8;
+		int height = recipe.getDisplayHeight() + 8;
+		EmiScreenshotRecorder.saveScreenshot("recipe-", width, height,
```
emilyploszaj, line 28: The format of saving name by timestamp is... probably not ideal for most cases. I think it'd be better to save the file based on the recipe's identifier if possible, defaulting to a system like this if null. It is worth noting that EMI recipe identifiers can represent illegal paths like starting with `/` or having `//` in the body, so that character should be replaced with something else. For organization's sake, it should probably be a subfolder inside screenshots as well.

src/main/java/dev/emi/emi/EmiScreenshotRecorder.java
```diff
@@ -0,0 +1,112 @@
...
+		if (EmiConfig.recipeScreenshotScale < 1) {
+			scale = EmiPort.getGuiScale(client);
+		} else {
+			scale = EmiConfig.recipeScreenshotScale;
+		}
+
+		Framebuffer framebuffer = new SimpleFramebuffer(width * scale, height * scale, true, MinecraftClient.IS_SYSTEM_MAC);
+		framebuffer.setClearColor(0f, 0f, 0f, 0f);
+		framebuffer.clear(MinecraftClient.IS_SYSTEM_MAC);
+
```
emilyploszaj, line 46: Initialization should contain a line somewhere invoking the following (as well as enabling blending) to prevent transparency issues discussed.
`RenderSystem.blendFuncSeparate(GLStateManager.SrcFactor.SRC_ALPHA, GLStateManager.DstFactor.ONE_MINUS_SRC_ALPHA, GLStateManager.SrcFactor.SRC_ALPHA, GLStateManager.DstFactor.ONE_MINUS_SRC_ALPHA);`

emilyploszaj (changes requested):
Overall looks pretty good, just a couple things I'd like to see resolved.

src/main/java/dev/emi/emi/EmiScreenshotRecorder.java
Kneelawk: I ended up fixing the fluid transparency issues by enabling blending and setting the blend func in the fluid stack renderer. See 743f41f.

src/main/java/dev/emi/emi/widget/RecipeScreenshotButtonWidget.java
Kneelawk: Addressed by dd3b1a8 I believe.

src/main/java/dev/emi/emi/screen/RecipeScreen.java
Kneelawk: Addressed by 9325093.

Kneelawk:
Ok, I've changed the existing things, next I'm going to test with the Create and Botania widgets.

emilyploszaj (approved):

Kneelawk:
Ok, it works with Botania now.

Comparison: old (first) vs. new (second)
mana_infusion_grass_seeds
mana_infusion_grass_seeds_3

PR #101, opened 2022-11-23
Add TextWidget alignment options

Kneelawk:
# This PR
This PR adds options to `TextWidget` to allow users to specify alignment.

Unlike most widgets, the user cannot necessarily know the dimensions of their text because different translations may change text length. If a user wanted to have text be aligned to the end or center of a recipe, they would have to calculate the width of their text themselves. This PR makes that calculation a part of `TextWidget`.

## Screenshot
Here is an example of if the furnace recipe had its Exp text aligned to the end of the recipe:
glass


src/main/java/dev/emi/emi/api/widget/Alignment.java
```diff
@@ -0,0 +1,13 @@
+package dev.emi.emi.api.widget;
+
+public enum Alignment {
```
emilyploszaj, line 3: This name is too basic and will conflict with other classes. If it were an inner class `TextWidget$Alignment`, that would be fine.

src/main/java/dev/emi/emi/api/widget/TextWidget.java
```diff
@@ -19,6 +21,16 @@ public TextWidget(OrderedText text, int x, int y, int color, boolean shadow) {
...
+		return this;
+	}
+
+	public TextWidget verticalAlign(Alignment alignment) {
+		this.verticalAlignment = alignment;
+		return this;
+	}
+
 	@Override
 	public Bounds getBounds() {
```
emilyploszaj, line 35: Bounds are not shifted based on alignment.

emilyploszaj (commented):
I find the presence of vertical alignment peculiar but I can't really justify excluding it.

src/main/java/dev/emi/emi/api/widget/Alignment.java
Kneelawk: Ok, makes sense.

Kneelawk:
Yeah, I guess vertical alignment is more of a QOL & uniformity thing.

src/main/java/dev/emi/emi/api/widget/TextWidget.java
```diff
@@ -5,11 +5,14 @@
 import net.minecraft.text.OrderedText;
 
 public class TextWidget extends Widget {
+	private static final int HEIGHT = 10;
```
emilyploszaj, line 8: What's the point of hardcoding height? We have access to it in the text renderer so we might as well use it if we support vertical alignment.

Kneelawk: It had been hard-coded previously so I had thought that was intentional.

`TextRenderer.fontHeight` is hard-coded as `9` but in `getBounds` height was hard-coded as 10, so I figured that was the way you wanted it.

I can change it to use `TextRenderer.fontHeight` if you want.

emilyploszaj: I probably literally just guessed, we should use fontHeight

Kneelawk: Ok, I'll do that then.

emilyploszaj (approved):

PR #136, opened 2023-02-01
Use ScreenHandler#getType method instead of using mixin accessor

deirn:
Fix issues with modded ScreenHandlers extending vanilla that overrides `getType` to return its own.


emilyploszaj (changes requested):
Going with a try catch approach could be concerning for performance. I'd prefer ever type of exception was caught and that on assignment you special case `screenHandler instanceof PlayerScreenHandler` to set to null and avoid throws optimistically.

deirn:
There are more screen handlers that have `null` type on vanilla though, like the creative and horse inventory.
Maybe mixin-ing `getType` to not throw an error when an injected field is true? Something like this:
```java
public ScreenHandlerType<?> getType() {
    if (this.type == null) {
        if (emi$bypassError) return this.type;
        throw new UnsupportedOperationException("Unable to construct this menu by type");
    }
    return this.type;
}
```

deirn:
Mixining ScreenHandler seems to work.

emilyploszaj:
Ah, I think this mixin is overkill here. While other screen handlers may use null, it's fine to catch the exception in that case, because they're not going to have recipe handlers, so performance (for updating craftables etc) is not going to be a concern. And, truthfully, it's just kinda a nasty mixin to have to maintain. I'd be fine with the initial implementation provided there's just an *extra* case in the try block that skips the call only if it's a `PlayerScreenHandler`, sorry if I didn't convey that properly.

emilyploszaj (approved):
Looks great now, sorry for the runaround!

ConqAra:
Will these changes fix integration with [slotlink](https://github.com/badasintended/slotlink/issues/152) once the new emi update comes out with the recent commits? Even with deirn's update to slotlink fixing it, I'm still unable to grab items from the storage with EMI interface enabled. I can use the Fill Recipe button now as intended, but the game still treats the storage as a recipe viewer, so I can never take anything out of it to put into my inventory.

PR #157, opened 2023-03-27
Only add stacks to the lookup history if recipes or usages are found

### LINKED ISSUE #153: Items are still added to the lookup history even when no recipes/usages are found

Abbie5:
Fixes #153


emilyploszaj:
I think this is a bit of a messy solution to the problem, perhaps the stack context could get passed along to `setPages`?

Abbie5:
Updated with passing the stack to setPages, for uses of setPages with no associated stack i went with EmiStack.EMPTY

emilyploszaj (approved):

PR #234, opened 2023-07-05
Handling For Horrible Horrible Contract-Breaking Non-PotionItem BrewingItemRecipes

sisby-folk:
Using brewing for non-potions! very bad idea! breaks implicit contracts of the game! except it only takes a [couple](https://github.com/HestiMae/pollinators-paradise/blob/1.20/src/main/java/garden/hestia/pollinators_paradise/mixin/PotionSlotMixin.java) of [mixins](https://github.com/HestiMae/pollinators-paradise/blob/1.20/src/main/java/garden/hestia/pollinators_paradise/mixin/BrewingRecipeRegistryMixin.java) to allow Brewing`itemRecipes`, which are usually used for transforming _any_ potion from one PotionItem to another, to be used for a [regular item->item transformation recipe](https://github.com/HestiMae/pollinators-paradise/blob/1.20/src/main/java/garden/hestia/pollinators_paradise/PollinatorsParadise.java#L67).

In a regular item->item recipe, the potion NBT being clipped onto the item does nothing - meaning you get 14 pages of identical recipe displays - but! If we check for the input being PotionItem _while_ generating displays, we can pass it just one potion instead if it's a funky little contract breaker.

We were going to just mixin/api into EMI for this one, but the idea was thrown around that this might be the desired behaviour in general for most mods breaking this contract - so hey why not PR it!

Tested (on fabric at least) and doesn't break gunpowder!

-- Clawby

### DIFF
--- a/fabric/src/main/java/dev/emi/emi/platform/fabric/EmiAgnosFabric.java
+++ b/fabric/src/main/java/dev/emi/emi/platform/fabric/EmiAgnosFabric.java
@@ -4,6 +4,7 @@ import java.nio.file.Path;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.Consumer;
 
 import dev.emi.emi.EmiPort;
 import dev.emi.emi.EmiRenderHelper;
@@ -34,12 +35,14 @@ import net.minecraft.client.util.math.MatrixStack;
 import net.minecraft.fluid.Fluid;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
+import net.minecraft.item.PotionItem;
 import net.minecraft.nbt.NbtCompound;
 import net.minecraft.potion.Potion;
 import net.minecraft.potion.PotionUtil;
 import net.minecraft.potion.Potions;
 import net.minecraft.recipe.BrewingRecipeRegistry;
 import net.minecraft.recipe.Ingredient;
+import net.minecraft.registry.entry.RegistryEntry;
 import net.minecraft.text.Text;
 import net.minecraft.util.Identifier;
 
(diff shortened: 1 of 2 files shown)


sisby-folk:
Oh also I know the consumer pattern is weird - initially just wanted to use a for loop with a conditional break on first success but that registry stream *really* doesn't like being anything but a stream so we did it this way.

emilyploszaj:
Definitely a weird consumer pattern but that's just how things like this go. I'm concerned with Forge, would it be possible to test to make sure Forge's brewing extensions don't create duplicate recipes and still works as intended? That's my only concern before a merge.

sisby-folk:
can do tomorrow - any good example mods? or should I just grab stuff randomly that's forge native

emilyploszaj:
I know Quark should have... something. Vazkii's Musketeer pack definitely has a couple and that's in public beta.

sisby-folk:
Tested with [potion of bees](https://github.com/Commoble/potionofbees) and [Extra Potions](https://github.com/CAS-ual-TY/Extra-Potions) on forge 1.20 if that means anything (both are forge native)

image
image

potion of bees did not show recipes, though it does use [it's own recipe type](https://github.com/Commoble/potionofbees/blob/main/src/main/java/commoble/potionofbees/PotionOfBeesRecipe.java#L11) and doesn't work on prod EMI either.

Not sure how to target/merge EMI branches for trying out 1.19.2 stuff like quark.

sisby-folk:
Got it on musketeer, Here's thorns splash:

image

The altered regen recipe doesn't work - it shows the splash->splash recipe, but not regular->splash.

image

image

This is also true in prod EMI though.

emilyploszaj (approved):

PR #319, opened 2023-09-28
Added mouse input handler and key input handle for JemiRecipe

Sleepwalkerx:
I don't know for what reason it wasn't implemented, but due to the absence of this code, JEI recipes don't handle clicks and keystrokes.

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/jemi/JemiRecipe.java
+++ b/xplat/src/main/java/dev/emi/emi/jemi/JemiRecipe.java
@@ -1,12 +1,6 @@
 package dev.emi.emi.jemi;
 
-import java.util.List;
-import java.util.Optional;
-
-import org.jetbrains.annotations.Nullable;
-
 import com.google.common.collect.Lists;
-
 import dev.emi.emi.EmiUtil;
 import dev.emi.emi.api.recipe.EmiRecipe;
 import dev.emi.emi.api.recipe.EmiRecipeCategory;
@@ -24,8 +18,14 @@ import mezz.jei.api.gui.IRecipeLayoutDrawable;
 import mezz.jei.api.recipe.RecipeIngredientRole;
 import mezz.jei.api.recipe.category.IRecipeCategory;
 import mezz.jei.library.focus.FocusGroup;
+import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.gui.tooltip.TooltipComponent;
+import net.minecraft.client.util.InputUtil;
 import net.minecraft.util.Identifier;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+import java.util.Optional;
 
 public class JemiRecipe<T> implements EmiRecipe {
 	public List<EmiIngredient> inputs = Lists.newArrayList();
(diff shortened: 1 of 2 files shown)


emilyploszaj:
It is a bit silly that it was never implemented. However, I don't really want to add and expose an API like this on DrawableWidget. It'd be fine to just create a new widget for JEMI input events that doesn't render anything.

Sleepwalkerx:
Well, I made this code based on how you made DrawableWidget. I saw that you were creating a Consumer and decided to do it in the same style.
OK, then you propose to completely abandon DrawableWidget and make a separate widget for Jemi, or you want to create two widgets, but one for rendering and the other for intercepting inputs?

Sleepwalkerx:
Just if I understand you correctly, you want to create two widgets on top of each other, only one will be with a render, and the other without.
I don't see the point in this when you can just move all the logic (including render and tooltips) to a new JemiWidget.

emilyploszaj:
I was suggesting leaving the drawable widget as it is, and adding a new widget, even just an anonymous or internal class, that just handles the input events.

Thinking about it though, a new custom widget class (inside JemiRecipe) that handles rendering, tooltips, and input events for the JEMI display would probably be reasonable.

Sleepwalkerx:
I did as you said. Check it out.

SiverDX:
https://github.com/emilyploszaj/emi/blob/54b0490d852eb5c64e8ea8bddbf8bac9ba7b908e/xplat/src/main/java/dev/emi/emi/jemi/JemiRecipe.java#L153-L160

why are  you creating a second `EmiDrawContext`

Sleepwalkerx:
> https://github.com/emilyploszaj/emi/blob/54b0490d852eb5c64e8ea8bddbf8bac9ba7b908e/xplat/src/main/java/dev/emi/emi/jemi/JemiRecipe.java#L153-L160
> 
> why are you creating a second `EmiDrawContext`

because it is done in the original logic. I don't want to break anything...

Sleepwalkerx:
I removed it.

xplat/src/main/java/dev/emi/emi/jemi/JemiRecipe.java
```diff
@@ -134,4 +129,55 @@ public void addWidgets(WidgetHolder widgets) {
...
+		@Override
+		public boolean mouseClicked(int mouseX, int mouseY, int button) {
+			return category.handleInput(recipe, mouseX, mouseY, InputUtil.Type.MOUSE.createFromCode(button));
+		}
+
+		@Override
+		public boolean keyPressed(int keyCode, int scanCode, int modifiers) {
+			MinecraftClient client = MinecraftClient.getInstance();
+			double mouseX = client.mouse.getX() * (double)client.getWindow().getScaledWidth() / (double)client.getWindow().getWidth();
+			double mouseY = client.mouse.getY() * (double)client.getWindow().getScaledHeight() / (double)client.getWindow().getHeight();
```
emilyploszaj: I'd prefer to use `EmiScreenManager.lastMouseX` and `EmiScreenManager.lastMouseY` just to reduce the use of calculations.

emilyploszaj (approved):
Other than the comment I have it looks good to me. Have you tested this with mods that use JEI's input features to make sure it works?

Sleepwalkerx:
> Other than the comment I have it looks good to me. Have you tested this with mods that use JEI's input features to make sure it works?

Initially, I discovered this problem when using the CustomMachinery mod and after these changes, the click in JEI began to work normally for CM. In general, I think it will work fine.

emilyploszaj (approved):

PR #320, opened 2023-09-29
Added searching by id option

### LINKED ISSUE #251: Suggestion: Add search by ID config option
Or something akin to this. It would be nice to search based on ID instead of name.
Use cases:
* I'm learning another language, but I don't know what to type in said language.
* I'm playing with Pirate Language turned on, but don't know that "Glass" is now "Porthole", so I can't find it.
* More generally, I only know the ID of an item, but not the name of it in the current locale.

kidofcubes:
Closes #251

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/config/EmiConfig.java
+++ b/xplat/src/main/java/dev/emi/emi/config/EmiConfig.java
@@ -56,6 +56,10 @@ public class EmiConfig {
 	public static HelpLevel helpLevel = HelpLevel.NORMAL;
 
 	@ConfigGroup("general.search")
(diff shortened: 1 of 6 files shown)


emilyploszaj (changes requested):
Thanks for the contribution! Unfortunately, the implementation is a bit unlike what I'd want.

One of the main concerns for searching by ID is mod id, rather than name. Personally, I think it'd be fine to just add both mod name and mod namespace to the mod search as well as item id path to the name lookup. I don't think we need another option for searching by this, if necessary, it can be added later.

emilyploszaj:
This has been implemented a few versions ago in the way I described, so I'm going to close this PR

PR #333, opened 2023-10-20
Use loose generics as much as possible

Phoupraw:
Before, you need to cast generics in some cases:
```java
List<EmiStack> all;
@Override
public List<EmiIngredient> getInputs() {
    return (List<EmiIngredient>) (List<? extends EmiIngredient>) all;
}
@Override
public List<EmiStack> getOutputs() {
    return all;
}
```
Now you can just return it directly:
```java
List<EmiStack> all;
@Override
public List<? extends EmiIngredient> getInputs() {
    return all;
}
@Override
public List<? extends EmiStack> getOutputs() {
    return all;
}
```

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/api/recipe/EmiInfoRecipe.java
+++ b/xplat/src/main/java/dev/emi/emi/api/recipe/EmiInfoRecipe.java
@@ -12,16 +12,17 @@ import net.minecraft.client.MinecraftClient;
 import net.minecraft.text.OrderedText;
 import net.minecraft.text.Text;
 import net.minecraft.util.Identifier;
+import org.jetbrains.annotations.Unmodifiable;
 
 public class EmiInfoRecipe implements EmiRecipe {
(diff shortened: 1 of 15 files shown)


emilyploszaj:
I appreciate the thought here but unfortunately this is a breaking API change, so it cannot be merged. In addition, the PR contains a handful of unrelated adjustments and changes, and some notable style violations like usage of `var` which is not something I use in my codebases. I don't think this PR can be merged even with changes, however, because of the status as a serious out of band API breakage.

As a justification for why the API is like it is: it's more convenient for callers, and for implementations, you're expected to just be constructing a list and storing it as a field, the use case of duplicate inputs and outputs is relatively niche, and the cast isn't that bad or common.

Sorry to close this immediately but all breaking changes need to be considered carefully and managed during an update cycle with concern to not cause undue churn for API consumers.

Phoupraw:
Oh no, I didn't realize it was a breaking update. Alas, I should have reacted when I encountered a large number of variable declaration compilation errors.

PR #337, opened 2023-11-01
Add support for Priority in the Resource Pack json files, used to edit Stack Visibility

Kanzaji:
This PR implements a Priority field in Resource Pack json files, used to sort `stackData` Suppliers in a way the highest priority will edit the Stack Visiblity last, giving it in effect highest Priority.

This feature is implemented to fix load order issues I've been mentioning for past month or two, that caused items added by `x.json` file being again hidden by `y.json` file, because it was loaded after the x file.

Priority defaults to value 0, what in result doesn't touch the order of files being loaded, not affecting already live Resource Packs.


emilyploszaj (changes requested):
I'm not sure why a list of *maps* is being used here, since the list type is what is returned by each json, in this case, a supplier of `IndexStackData`. Instead of having priority be a key for the map, there could be a `PrioritySupplier` type which has both a `get` like a supplier and a `getPriority` for the priority, which could be added to the list instead of the `Supplier`s. At the end, a simple sort over the list could be done. `PrioritySupplier` could be a `public static class` inside `EmiData`. This would let this sorting system be used for other data types easily with minimal changes and be much more reasonable than a list of maps.

Kanzaji:
@emilyploszaj Reworked the PR with the feedback I've got, I hope this is better ❤️

emilyploszaj (approved):
This is exactly what I was imagining, this looks great.

PR #367, opened 2023-12-09
Update 'russian' translation

eterna1-0blivion:
- The order is overridden according to the position on the settings screen (key registration).
- Comments have been added to the original file to ease third-party translation (it's a good tone).
- Added positions for changing the description on the mod menu screen (at the beginning of the file).
- Many positions in the Russian translation have been revised. Tooltips, items, tags and other. There is no limit to perfection, but it is definitely better.
- Found an redundant item "key.emi.set_default" at line 245 - tagged with question marks.


emilyploszaj:
English tooltips for the config are intentional left untranslated. The translations here seem to additionally differ than what currently exists. The resort makes it pretty difficult to reasonably audit. I'm willing to merge the russian updates but you'll need to revert the changes to english.

xplat/src/main/resources/assets/emi/lang/ru_ru.json
```diff
@@ -291,70 +522,70 @@
...
-  "tag.item.minecraft.soul_fire_base_blocks": "Основы для огня душ",
-  "tag.item.minecraft.stairs": "Ступени",
-  "tag.item.minecraft.stone_bricks": "Каменные кирпичи",
-  "tag.item.minecraft.stone_crafting_materials": "Каменные материалы",
-  "tag.item.minecraft.stone_tool_materials": "Каменные инструменты",
-  "tag.item.minecraft.tall_flowers": "Высокие цветы",
-  "tag.item.minecraft.terracotta": "Терракота",
-  "tag.item.minecraft.trapdoors": "Люки",
-  "tag.item.minecraft.walls": "Стены",
+  "tag.minecraft.fishes": "Рыба",
```
emilyploszaj: There's a collection of translations here that replace `tag.item.minecraft` with `tag.minecraft`. This should not be the case.

emilyploszaj (changes requested):
I appreciate the translation. As noted, the changes to english need to be reverted. A more widespread misconception is the use of question marks in tooltips, the phrases are not "Where to display status effects in the inventory?" it's the implied "{[The setting] is} where to display status effects in the inventory" construction. This may make less sense in other languages, but it's frequent here and I assume it may be a misinterpretation.

eterna1-0blivion:
> English tooltips for the config are intentional left untranslated. The translations here seem to additionally differ than what currently exists. The resort makes it pretty difficult to reasonably audit. I'm willing to merge the russian updates but you'll need to revert the changes to english.

Got the English translation back

xplat/src/main/resources/assets/emi/lang/ru_ru.json
eterna1-0blivion: I think it's just a mistake. I'll fix it.

eterna1-0blivion:
> I appreciate the translation. As noted, the changes to english need to be reverted. A more widespread misconception is the use of question marks in tooltips, the phrases are not "Where to display status effects in the inventory?" it's the implied "{[The setting] is} where to display status effects in the inventory" construction. This may make less sense in other languages, but it's frequent here and I assume it may be a misinterpretation.

In this case, the question in Russian sounds good and logical quite well. I have enough experience in writing texts of various kinds to say this with confidence. Maybe I'll come back to check the translation in a week to take a fresh look at it, but I'm convinced that my version is already much better than the existing one for two reasons:
- There were obvious logical errors in the names of the settings, because it's almost impossible to translate them without context from English
- And this leads to the second reason - translation of the tooltips, which were also very difficult to translate because of their double interpretation in some cases.

I will definitely need to play around with this translation to fix minor bugs and mistakes, but for now I'm happy with it.

xplat/src/main/resources/assets/emi/lang/en_us.json
```diff
@@ -702,4 +702,4 @@
   "tag.item.fabric.hoes": "Hoes",
   "tag.item.fabric.swords": "Swords",
   "tag.item.fabric.shears": "Shears"
-}
+}
```
emilyploszaj: Nit: there's still a change here, removing of a blank line

emilyploszaj (approved):
With your confidence and a quick pass as best as I can, this looks reasonable to merge once the `en_us.json` single line change is reverted. Thanks for your contribution.

xplat/src/main/resources/assets/emi/lang/en_us.json
eterna1-0blivion: Done.

PR #387, opened 2023-12-25
Add Minecraft Recipe IDs where appropriate

shartte:
To aid mods that need to correlate EMI recipes with Minecraft recipes (i.e. for saving references to them in NBT data or similar), this adds a defaulted method to access the Minecraft Recipe ID via EmiRecipe.
For many of the EMI recipes, this is just the normal recipe id (but not for all...).


emilyploszaj:
Since this is now in EMI this PR is no longer necessary

PR #399, opened 2024-01-04
Mod name tooltips: match kebab case mod IDs, fall back to friendly namespaces

sisby-folk:
Attempts to get a mod container with ID `cool-mod` if the container with ID `cool_mod` doesn't exist.

If neither exist, falls back to `Cool Mod` instead of `cool_mod` using basic string formatting. 

Users that need IDs for commands can use F3+H.

### DIFF
--- a/fabric/src/main/java/dev/emi/emi/platform/fabric/EmiAgnosFabric.java
+++ b/fabric/src/main/java/dev/emi/emi/platform/fabric/EmiAgnosFabric.java
@@ -45,6 +45,7 @@ import net.minecraft.recipe.Ingredient;
 import net.minecraft.registry.entry.RegistryEntry;
 import net.minecraft.text.Text;
 import net.minecraft.util.Identifier;
+import org.apache.commons.lang3.text.WordUtils;
 
 public class EmiAgnosFabric extends EmiAgnos {
 	static {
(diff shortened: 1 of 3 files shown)


sisby-folk:
example with `loggerhead-luminancies`

release:

image

pr:

image

emilyploszaj:
I can understand prettying up IDs, particularly good for stuff like kubejs items, tags, and recipes, but the _ to - conversion does have me a bit hesitant, that seems like something modders should be handling. Alternatively, having a translation key for mod names could be a reasonable solution as well.

sisby-folk:
Yea to give our use case - we like kebab case everywhere - mod slugs, project name, mod ID, archive name, etc - we think it looks nice and actively decide to not use snake.

But kebab is illegal for packages & therefore effectively namespace. Having to change our entire published mod ID in an update because we added an item has been a consideration due to this limitation before and I reckon we can just do better via this.

emilyploszaj:
Yeah I'm swayed, though I believe WordUtils is deprecated and would enjoy that warning be suppressed or an alternative method be used.

sisby-folk:
As far as I'm aware there's not a direct alternative available on the
minecraft classpath, so i guess suppression is best?

emilyploszaj:
Yeah, just chuck in a suppression then, it's a losing battle but I try to keep as few warnings in my workspace as possible

PR #418, opened 2024-01-24
allow category properties to set identifier sorting

sisby-folk:
Identifier sorting exists but you can't set it via assets, so this fixes that. Adding that info to the table in https://github.com/emilyploszaj/emi/wiki/Recipe-Category-Properties would be good too.


emilyploszaj:
Unclear why I never merged this, this looks fine.

PR #452, opened 2024-03-04
Make the EMI Server-Side component optional on Neoforge

shartte:
I forgot to make the registrar optional, making the server-side component required on Neoforge. With this change, I was able to successfully connect to a Server that doesn't have EMI with just the client-side mod installed.

Also update to stable 1.20.4 Neoforge version.


emilyploszaj:
Uh oh, that's not ideal, thanks for the fix. Hopefully I can get a chance to release this soon.

PR #455, opened 2024-03-07
Indicate on EmiStackInteraction that recipe is nullable

shartte:
While unrelated to this change, I was bitten by EmiStackInteraction's shorter constructor setting clickable to `true` by default, since it overrides my slot default behavior :-)

In any case, since I now know that I need to use the full constructor to set clickable to false, I noticed that recipe is not actually marked as nullable, even though it is.


emilyploszaj:
I'm relatively inconsistent on nullability annotation but this seems like one I definitely should've had.

PR #474, opened 2024-03-16
updated lang files for ja_jp (1.20.4)

medi-torimorta:
updated to be more accurate and sound natural to Japanese audiences


emilyploszaj (approved):
This looks reasonable but was a bit difficult to review due to the entire file's line endings or similar being changed, resulting in an unclean diff. I'd prefer that be avoided in the future if possible.

PR #479, opened 2024-03-19
Allow the batching renderer to be used with new enough Sodium versions

embeddedt:
Modern versions of Sodium (0.5.4+) correctly fall back to using `VertexConsumer` even if it doesn't implement the `VertexBufferWriter` interface. Therefore, hard-disabling the batcher should no longer be required. This PR adjusts the detection logic to detect the specific version(s) of Sodium that would have this problem and only disables the batcher for those specific versions.

I've tested that the batcher does work fine with Sodium 0.5.8 on MC 1.20.4, unfortunately there is no Sodium 0.4 version to test with on this Minecraft version.

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/screen/StackBatcher.java
+++ b/xplat/src/main/java/dev/emi/emi/screen/StackBatcher.java
@@ -45,8 +45,8 @@ import net.minecraft.item.ItemStack;
  */
 public class StackBatcher {
 	private static MethodHandle sodiumSpriteHandle;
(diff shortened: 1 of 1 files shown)


emilyploszaj:
Could you merge this commit onto an older version branch locally to verify this behaves correctly? There should be no merge conflicts.

embeddedt:
Sure. Checked on 1.20.1, it correctly disables the batching renderer when running alongside Sodium 0.4.10.

Issue #469, opened 2024-03-15
Missing sources from the source jar


pupnewfster:
When trying to hit download sources with Intellij for EMI version 1.1.3 neoforge 1.20.4 it fails, and from trying to manually attach the sources I believe that is because the majority of directories (especially the api related ones) seem to be empty. I downloaded the source jar from https://maven.terraformersmc.com/releases/dev/emi/emi-neoforge/1.1.3+1.20.4 and viewing it with a zip viewer and navigating to `dev/emi/emi` shows this for the sizes of different directories.
![image](https://github.com/emilyploszaj/emi/assets/1203288/50bcf036-ec9b-4b22-8adc-e14df5e7e0ea)

Is there some way this could be fixed, as it makes getting used to a new API and figuring things out quite difficult when having to browse the source files on github to see the javadocs rather than being able to see them natively in my IDE.

emilyploszaj:
Wonderful, this error again...

Loom, and consequentially Arch Loom, have a peculiar remapping issue where certain symbols cause an NPE when remapping sources that will silently fail and print no debuggable information. It'll also stop the remapping process wherever it was, and since the order in which it iterates classes and folders is random, it's really hard to tell what actually causes it. Resolving this is basically trial and error, and will likely take some time. Furthermore, resolving this actually involves manually excluding files from the sources, so it's a "pick your poison" moment of choosing which files to be excluded instead of having an arbitrary amount go missing. From a quick look, the fabric artifact seems fine still, so this is something relating to Neo specifically.

I looked at the latest handful of releases, and 1.1.1 seems to have the most complete sources, having at least all of the api package and subpackages mapped. As a mod dev I don't think you'll suffer any consequences while this is being resolved if you depend on that slightly older version. Hopefully the next release has this fixed, but depending on release timelines it might take 2 versions. Thanks for pointing this out.

pupnewfster:
Nice thanks, using 1.1.1 does seem to allow the sources to attach and allows me to view the docs from my IDE

pupnewfster:
I did some investigating locally, and it seems to be because the remap fails as JEI isn't on the classpath in neoforge, and is only on it for fabric, so when trying to remap a jemi file it crashes and stops remapping. Exlcluding jemi from the source jar for neo rather than just the filtered source jar seems to allow it to remap properly

emilyploszaj:
Fixed in EMI 1.1.4 after some gradle wrestling, thanks for the PR.

PR #486, opened 2024-03-22
Cleanup build.gradle files

pupnewfster:
- Make use of lazy configuration which is the way of defining things that gradle has been slowly moving towards
- Fixes all gradle 9 warnings that are not just warnings of architectury loom as a plugin
- Fix source jars for neo not generating properly (#469). I am effectively just making both source jars for neo be filtered now instead of just the filtered one, as the remapper seems to crash when it encounters JEMI classes as JEI isn't on the classpath

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,7 @@ allprojects {
 		compileOnly()
 	}
 
-	archivesBaseName = rootProject.archives_base_name
+	base.archivesName = rootProject.archives_base_name
 	ext.mavenVersion = "${rootProject.mod_version}"
 	if (System.getenv("RELEASE") == null) {
 		ext.mavenVersion += "-SNAPSHOT"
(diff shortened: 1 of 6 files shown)


emilyploszaj:
Modern Gradle sure looks a lot worse than it used to... Did you check that the artifacts generated from the various loaders all look correct (or, well, better, in the case of neo)? I generally don't like touching Gradle since I have so many different projects and artifacts to screen, heh. Hopefully I can merge these changes back to 1.19.2, I'll have to take a look and see if the build system is gathering tech debt.

pupnewfster:
I published it to maven local and am building and running against the neo one in dev, and can properly see the sources for it there.

emilyploszaj (approved):
Guess I'll have to do a deeper dive then for other platforms, thanks for the contribution!

emilyploszaj:
https://github.com/emilyploszaj/emi/actions/runs/8398429595/job/23003265915 Ah, this has broken my release system. Do you know any quick resolution to this?

emilyploszaj:
Was able to resolve it, gradle's always headaches it seems.

pupnewfster:
Sorry, I missed it, I guess I didn't look closely enough at the curseforge/modrinth blocks in part given the plugins weren't actually loaded as I didn't have keys for them set.

PR #496, opened 2024-03-30
Add an API to allow mods to do their own EMI searching

embeddedt:
This PR aims to add an API so that mods can run their own EMI search queries without changing the contents of the global search bar. (The motivation for this change is to add support for EMI to ModernFix's creative search tree optimization, which skips vanilla's search tree creation and just uses the installed recipe viewer to do the searching instead.)

To do this, some static fields were refactored into a "search manager" object. EMI's screen manager holds one instance of such an object, and mods may create their own via `EmiApi.createSearchManager()`.

The code may need some quality improvement work, and an outstanding question is whether some guard needs to be added to stall searches until `EmiSearch.bake()` has finished running.

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/api/search/EmiSearchManagerApi.java
+++ b/xplat/src/main/java/dev/emi/emi/api/search/EmiSearchManagerApi.java
@@ -1,0 +1,24 @@
+package dev.emi.emi.api.search;
+
+import dev.emi.emi.api.stack.EmiIngredient;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * A search manager controls searching for items using EMI infrastructure.
+ */
+public interface EmiSearchManagerApi {
+    /**
+     * {@return the current list of stacks matching the last search query}
+     */
+    List<? extends EmiIngredient> getStacks();
+
+    /**
+     * Search for ingredients matching the given query string. The list returned by {@link EmiSearchManagerApi#getStacks()}
+     * will not update until after the returned future completes.
+     * @param query the query string to use when searching
+     * @return a future that completes with the updated list
+     */
+    CompletableFuture<List<? extends EmiIngredient>> search(String query);
+}
--- a/xplat/src/main/java/dev/emi/emi/search/EmiSearchManager.java
+++ b/xplat/src/main/java/dev/emi/emi/search/EmiSearchManager.java
@@ -1,0 +1,98 @@
+package dev.emi.emi.search;
+
+import com.google.common.collect.Lists;
+import dev.emi.emi.api.search.EmiSearchManagerApi;
+import dev.emi.emi.api.stack.EmiIngredient;
+import dev.emi.emi.api.stack.EmiStack;
+import dev.emi.emi.registry.EmiStackList;
+import dev.emi.emi.runtime.EmiLog;
+import dev.emi.emi.screen.EmiScreenManager;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+public class EmiSearchManager implements EmiSearchManagerApi {
+    private EmiSearch.CompiledQuery compiledQuery;
+    private List<? extends EmiIngredient> stacks = EmiStackList.stacks;
+    private volatile SearchWorker currentWorker;
+
+    public CompletableFuture<List<? extends EmiIngredient>> search(String query) {
+        synchronized (this) {
+            SearchWorker worker = new SearchWorker(query, EmiScreenManager.getSearchSource());
+            currentWorker = worker;
+
+            EmiSearch.executor.execute(worker);
+            return worker.getCompletionFuture();
+        }
+    }
+
+    public List<? extends EmiIngredient> getStacks() {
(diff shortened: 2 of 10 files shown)


xplat/src/main/java/dev/emi/emi/api/search/EmiSearchManagerApi.java
```diff
@@ -0,0 +1,24 @@
...
+
+import dev.emi.emi.api.stack.EmiIngredient;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * A search manager controls searching for items using EMI infrastructure.
+ */
+public interface EmiSearchManagerApi {
```
emilyploszaj: I prefer an `Foo` `FooImpl` styling compared to `FooApi` `Foo`

xplat/src/main/java/dev/emi/emi/search/EmiSearchManager.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.concurrent.CompletableFuture;
+
+public class EmiSearchManager implements EmiSearchManagerApi {
+    private EmiSearch.CompiledQuery compiledQuery;
+    private List<? extends EmiIngredient> stacks = EmiStackList.stacks;
+    private volatile SearchWorker currentWorker;
+
+    public CompletableFuture<List<? extends EmiIngredient>> search(String query) {
+        synchronized (this) {
+            SearchWorker worker = new SearchWorker(query, EmiScreenManager.getSearchSource());
```
emilyploszaj: This doesn't work like you imagine. EMI's search source is, currently (to be adjusted), always the current contents of the right sidebar. This means that searching will be dependent on the user's current setup. There should be a way for EMI's needs to set this to the current search source, but have the API exposed methods always use the index's stack list.

xplat/src/main/java/dev/emi/emi/search/EmiSearchManager.java
```diff
@@ -0,0 +1,98 @@
...
+                    apply(source);
+                    return;
+                }
+                List<EmiIngredient> stacks = Lists.newArrayList();
+                int processed = 0;
+                for (EmiIngredient stack : source) {
+                    if (processed++ >= 1024) {
+                        processed = 0;
+                        if (this != currentWorker) {
+                            apply(source);
```
emilyploszaj: Is this desirable, completing a partially finished query? Should this behavior even exist for API searching?

xplat/src/main/java/dev/emi/emi/api/search/EmiSearchManagerApi.java
```diff
@@ -0,0 +1,24 @@
+package dev.emi.emi.api.search;
+
+import dev.emi.emi.api.stack.EmiIngredient;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * A search manager controls searching for items using EMI infrastructure.
```
emilyploszaj: Usage of "items" here in documentation is incorrect, EMI deals with "stacks" which are only often items.

xplat/src/main/java/dev/emi/emi/api/search/EmiSearchManagerApi.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+public interface EmiSearchManagerApi {
+    /**
+     * {@return the current list of stacks matching the last search query}
+     */
+    List<? extends EmiIngredient> getStacks();
```
emilyploszaj: Is it reasonable to expose a method that returns a `CompletableFuture` and a stateful method? This feels like confusingly designed API bloat.

emilyploszaj (changes requested):
I haven't done as thorough a review as I'd like as I do not currently have the time, but I have gone through and pointed out some initial things I noticed.

xplat/src/main/java/dev/emi/emi/search/EmiSearchManager.java
embeddedt: This has been addressed in the latest commit.

xplat/src/main/java/dev/emi/emi/search/EmiSearchManager.java
embeddedt: I've refactored it to not use the magic `currentWorker` global and instead support a proper method for cancelling, so the API consumer can decide what they want. EMI uses the cancelling to implement search the way it worked before.

xplat/src/main/java/dev/emi/emi/api/search/EmiSearchManagerApi.java
embeddedt: Got rid of it and moved to a cleaner `Future`-based interface for everything.

SergioJuniorCE:
is this still being worked on?

emilyploszaj:
It has been shelved a bit due to a lack of need on modern versions. It may be revisited when there is more time

VaporeonScripts:
is this being worked on? it would be cool to have that

PR #518, opened 2024-04-18
Port to 1.20.5

embeddedt:
Almost certainly contains bugs. However, it gets into the game, and some special recipes I checked manually (potions, anvil) seem to be working, so I believe it's correct for the most part.


fabric/src/main/java/dev/emi/emi/mixin/ClientConfigurationNetworkHandlerMixin.java
```diff
@@ -0,0 +1,16 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(ClientConfigurationNetworkHandler.class)
+public class ClientConfigurationNetworkHandlerMixin {
+    @Inject(at = @At("RETURN"), method = "onSynchronizeTags")
+    private void refreshTagBasedData(CallbackInfo info) {
+        EmiReloadManager.reloadTags();
```
emilyploszaj, line 14: Why are there two mixins that call `EmiReloadManager.reloadTags()`? EMI reloads once both tags and recipes are received and if multiple are called or in peculiar orders it may cause improper reloading behavior.

fabric/src/main/java/dev/emi/emi/mixin/SynchronizeRecipesS2CPacketMixin.java
```diff
@@ -15,13 +15,5 @@
...
-	
-	@Inject(at = @At("RETURN"), method = "readRecipe", locals = LocalCapture.CAPTURE_FAILHARD)
-	private static void readRecipe(PacketByteBuf buf, CallbackInfoReturnable<RecipeEntry<Recipe<?>>> info, Identifier serializer, Identifier id) {
-		RecipeEntry<Recipe<?>> recipe = info.getReturnValue();
-		if (recipe == null) {
-			EmiLog.error("Recipe " + id + " was deserialized as null using recipe serializer " + serializer
-				+ ", this is breaking vanilla recipe sync, and will prevent EMI from loading recipes");
-		}
-	}
+
```
emilyploszaj: Why is this class empty?

fabric/src/main/java/dev/emi/emi/mixin/api/EmiStackMixin.java
```diff
@@ -24,12 +24,12 @@ private static EmiStack of(ItemVariant item, long amount) {
 
 	@Transform(visibility = "PUBLIC")
 	private static EmiStack of(FluidVariant fluid) {
-		return new FluidEmiStack(fluid.getFluid(), fluid.getNbt());
```
emilyploszaj: anything in the `dev.emi.emi.mixin.api` package can be deleted without consequence, frankly all of the classes should probably be deleted in 1.20 and up, but deleting any errant classes here is not a problem.

fabric/src/main/java/dev/emi/emi/platform/fabric/EmiAgnosFabric.java
```diff
@@ -130,30 +134,26 @@ protected void addBrewingRecipesAgnos(EmiRegistry registry) {
...
+				Ingredient recipeIngredient = recipe.ingredient();
 				if (recipeIngredient.getMatchingStacks().length > 0) {
 					String gid = EmiUtil.subId(recipeIngredient.getMatchingStacks()[0].getItem());
-					String iid = EmiUtil.subId((Item) ((BrewingRecipeRegistryRecipeAccessor) recipe).getInput());
-					String oid = EmiUtil.subId((Item) ((BrewingRecipeRegistryRecipeAccessor) recipe).getOutput());
+					String iid = EmiUtil.subId(recipe.from().value());
+					String oid = EmiUtil.subId(recipe.to().value());
 					Consumer<RegistryEntry<Potion>> potionRecipeGen = entry -> {
-						Potion potion = entry.value();
-						if (potion == Potions.EMPTY) {
```
emilyploszaj, line 142: Why is this check removed?

gradle.properties
```diff
@@ -1,18 +1,18 @@
 org.gradle.jvmargs=-Xmx2048M
 
-minecraft_version=1.20.4
-enabled_platforms=fabric,neoforge
```
emilyploszaj, line 4: Does neo really not have any rc builds out yet?

xplat/src/main/java/dev/emi/emi/EmiPort.java
```diff
@@ -101,9 +104,10 @@ public static InputStream getInputStream(Resource resource) {
 		}
 	}
 
-	public static BannerPattern.Patterns addRandomBanner(BannerPattern.Patterns patterns, Random random) {
-		return patterns.add(Registries.BANNER_PATTERN.getEntry(random.nextInt(Registries.BANNER_PATTERN.size())).get(),
-			DyeColor.values()[random.nextInt(DyeColor.values().length)]);
+	public static BannerPatternsComponent addRandomBanner(BannerPatternsComponent patterns, Random random, DynamicRegistryManager registryManager) {
```
emilyploszaj: EmiPort exists to reduce commit diffs, changing the signature here makes this goal useless. The dynamic registry manager should just be gotten implicitly inside this method, and the added method does not seem to be used elsewhere, so it can be removed.

xplat/src/main/java/dev/emi/emi/EmiUtil.java
```diff
@@ -182,4 +185,9 @@ public static EmiRecipe getRecipeResolution(EmiIngredient ingredient, EmiPlayerI
 		}
 		return null;
 	}
+
+	public static ItemStack setPotion(ItemStack stack, RegistryEntry<Potion> potion) {
```
emilyploszaj: This is likely more applicable for EmiPort and would need to be a separate base commit.

xplat/src/main/java/dev/emi/emi/VanillaPlugin.java
```diff
@@ -327,6 +328,8 @@ public void register(EmiRegistry registry) {
 			EmiPort.getDisabledItems()
 		).collect(Collectors.toSet());
 
+		List<Item> dyeableItems = EmiUtil.values(ItemTags.DYEABLE).map(RegistryEntry::value).collect(Collectors.toList());
```
emilyploszaj, line 335: This probably makes more sense as a base abstraction rather than a version diff.

emilyploszaj: Additionally, is it reasonable to go based on the item tag rather than iterate the registry and collect everything that supports this interface/component?

xplat/src/main/java/dev/emi/emi/VanillaPlugin.java
```diff
@@ -639,7 +640,7 @@ private static void addWorldInteraction(EmiRegistry registry, Set<Item> hiddenIt
 			addRecipeSafe(registry, () -> basicWorld(EmiStack.of(entry.getKey()), honeycomb, EmiStack.of(entry.getValue()), id, false));
 		}
 
-		for (Item i : EmiArmorDyeRecipe.DYEABLE_ITEMS) {
+		for (Item i : EmiUtil.values(ItemTags.DYEABLE).map(RegistryEntry::value).collect(Collectors.toList())) {
```
emilyploszaj: This should probably be passed in from earlier, not ad hoc duplicate construction.

xplat/src/main/java/dev/emi/emi/api/stack/Comparison.java
```diff
@@ -8,7 +8,7 @@
 import dev.emi.emi.runtime.EmiLog;
 
 public class Comparison {
-	private static final Comparison COMPARE_NBT = Comparison.compareData(stack -> stack.getNbt()); 
+	private static final Comparison COMPARE_NBT = Comparison.compareData(stack -> stack.getComponents());
```
emilyploszaj: This constant and the method that returns it should probably be called "compare components". Any uses of this in EMI imply should probably additionally use an abstraction method to avoid version diffs.

xplat/src/main/java/dev/emi/emi/api/stack/EmiStack.java
```diff
@@ -23,7 +26,7 @@
  * An abstract representation of a resource in EMI.
  * Can be an item, a fluid, or something else.
  */
-public abstract class EmiStack implements EmiIngredient {
+public abstract class EmiStack implements EmiIngredient, ComponentHolder {
```
emilyploszaj: What are the implications of implementing `ComponentHolder`?

xplat/src/main/java/dev/emi/emi/api/stack/ItemEmiStack.java
```diff
@@ -137,7 +144,7 @@ public void renderForBatch(VertexConsumerProvider vcp, DrawContext draw, int x,
 
 	@Override
 	public List<Text> getTooltipText() {
-		return getItemStack().getTooltip(client.player, TooltipContext.BASIC);
+		return getItemStack().getTooltip(Item.TooltipContext.DEFAULT, client.player, TooltipType.BASIC);
```
emilyploszaj, line 161: Might be time to bite the bullet and add an `EmiPort.getTooltip(ItemStack)`

xplat/src/main/java/dev/emi/emi/jemi/JemiUtil.java
```diff
@@ -58,7 +58,8 @@ public static Optional<ITypedIngredient<?>> getTyped(EmiStack stack) {
 		if (stack.isEmpty()) {
 			return Optional.empty();
 		} else if (stack.getKey() instanceof Fluid f) {
-			return JemiPlugin.runtime.getIngredientManager().createTypedIngredient(getFluidType(), getFluidHelper().create(f, stack.getAmount() == 0 ? 1000 : stack.getAmount(), stack.getNbt()));
+			throw new UnsupportedOperationException("TODO port to components");
```
emilyploszaj: I'd prefer a more elegant error message like "Update JEMI integration with new JEI implementation" and more soft failure rather than hard failure.

xplat/src/main/java/dev/emi/emi/jemi/impl/JemiIngredientAcceptor.java
```diff
@@ -113,7 +113,8 @@ public JemiIngredientAcceptor addFluidStack(Fluid fluid, long amount) {
 
 	@Override
 	public JemiIngredientAcceptor addFluidStack(Fluid fluid, long amount, NbtCompound tag) {
-		addStack(EmiStack.of(fluid, tag, amount));
+		if(true) throw new UnsupportedOperationException("TODO port to components");
```
emilyploszaj: Unrelated to earlier exception comments, `if(true)` is a style violation and code smell here

xplat/src/main/java/dev/emi/emi/mixin/HandledScreenMixin.java
```diff
@@ -25,7 +25,7 @@ public abstract class HandledScreenMixin extends Screen {
 
 	@Dynamic
 	@Inject(at = @At(value = "INVOKE",
-			target = "net/minecraft/client/gui/screen/Screen.renderBackground(Lnet/minecraft/client/gui/DrawContext;IIF)V",
+			target = "Lnet/minecraft/client/gui/screen/ingame/HandledScreen;drawBackground(Lnet/minecraft/client/gui/DrawContext;FII)V",
```
emilyploszaj: This is a malformed invoke descriptor. Mixin will accept it, but for consistency my style uses proper formatting with `.` and no `L`.

xplat/src/main/java/dev/emi/emi/recipe/special/EmiArmorDyeRecipe.java
```diff
@@ -7,22 +7,19 @@
...
+import net.minecraft.component.type.DyedColorComponent;
 import net.minecraft.item.DyeItem;
-import net.minecraft.item.DyeableItem;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 import net.minecraft.util.DyeColor;
 import net.minecraft.util.Identifier;
 
 public class EmiArmorDyeRecipe extends EmiPatternCraftingRecipe {
-	public static final List<Item> DYEABLE_ITEMS = EmiPort.getItemRegistry().stream()
```
emilyploszaj: This removal can likely be abstracted.

emilyploszaj (commented):
Thanks, this is mostly looking good. Especially all the generated recipes, that looks like a pain.

Before merging this I should abstract some things in the general codebase to make merges more painless. The current list is:

Probably remove all self API mixins in later versions for sanity reasons.
`EmiPort.setPotion` similar to your `EmiUtil.setPotion`
`Comparison.compareNbt` should be phased out of impl usage because of the version name change, using an `EmiPort` method instead.
`EmiPort.getBasicTooltip(ItemStack)`
`EmiPort.isDamageable(Item)`
`EmiPort.canCombine(ItemStack, ItemStack)`

If you think anything should be added it'd be good to know, in the meantime I'll work on these changes so we can rebase and get this merged to a new branch.

emilyploszaj:
I believe I've done a majority of the abstractions, actually decided against a few I listed in the review comment because of how few usages there were or complexity with the abstraction. This should be a more reasonable base to work against.

fabric/src/main/java/dev/emi/emi/mixin/ClientConfigurationNetworkHandlerMixin.java
embeddedt: Tags now seem to be synchronized in the configuration listener as well as the play listener; without having the mixin here, it doesn't detect tags being synced at all on initial load.

fabric/src/main/java/dev/emi/emi/platform/fabric/EmiAgnosFabric.java
embeddedt: `Potions.EMPTY` no longer exists.

gradle.properties
embeddedt: Well, *now* we do, but it wasn't ready when I ported this yesterday. :laughing: I'll look into that.

xplat/src/main/java/dev/emi/emi/VanillaPlugin.java
embeddedt: > is it reasonable to go based on the item tag rather than iterate the registry and collect everything that supports this interface/component?

I don't think iterating the registry can work, because (from what I can tell) items do not define a set of supported components, only a set of default components (and I don't think the dyeable items necessarily have the dye component attached by default).

xplat/src/main/java/dev/emi/emi/api/stack/EmiStack.java
embeddedt: `ComponentHolder` just defines `ComponentMap getComponents()`, the rest of the methods are default wrappers that delegate to the component map. Main reason I implemented it was for "nice" QoL methods like `getOrDefault`.

xplat/src/main/java/dev/emi/emi/mixin/HandledScreenMixin.java
embeddedt: Really? Almost all mods I have seen (and the MinecraftDev plugin, which has had contributions from people who developed Mixin in the past) generate the target descriptors this way. Interested to know more.

fabric/src/main/java/dev/emi/emi/mixin/SynchronizeRecipesS2CPacketMixin.java
embeddedt: I'm not sure the patch can still apply easily to 1.20.5; the recipe packet appears to have changed and now uses packet codecs. Wanted to leave the class here so that it would be more obvious during review. Not sure what you want to do with it.

emilyploszaj: Yeah the class can be deleted then

xplat/src/main/java/dev/emi/emi/VanillaPlugin.java
emilyploszaj: Ah that makes sense, well then the tag will have to do

xplat/src/main/java/dev/emi/emi/api/stack/EmiStack.java
emilyploszaj: Good to know, I agree

xplat/src/main/java/dev/emi/emi/api/stack/ItemEmiStack.java
emilyploszaj: Decided this was too much of a pain

xplat/src/main/java/dev/emi/emi/mixin/HandledScreenMixin.java
emilyploszaj: mcdev is not an official plugin and it just so happens to use the "less correct" method of specifying descriptors. In the JVM constant pool, specifying a "type" is redundantly broad, because methods will always be called on classes, invokes are dot separated class/descriptor pairs. Technically, it'd be most correct to use `foo/Bar.baz:()V` with a colon but I'm both not sure if mixin supports that and honestly I don't think it looks good, plus it's non-ambiguous without it. The reason most people do it the wrong way is because, well, most people did it the wrong way and copy paste is how we survive as a society. It is simply more convenient to teach people "when you type in a class, always surround it with an L and a ; for some reason" rather than making the distinction between method invokes type usage. I have a weird bone to pick with this and it saves me one precious utf8 byte :P

fabric/src/main/java/dev/emi/emi/mixin/ClientConfigurationNetworkHandlerMixin.java
emilyploszaj: I see, I'll keep this open as a reminder for me to take a deeper look into it after I merge.

xplat/src/main/java/dev/emi/emi/api/stack/serializer/EmiStackSerializer.java
```diff
@@ -10,15 +10,17 @@
...
+import net.minecraft.nbt.NbtOps;
 import net.minecraft.nbt.StringNbtReader;
 import net.minecraft.util.Identifier;
 import net.minecraft.util.JsonHelper;
 
 public interface EmiStackSerializer<T extends EmiStack> extends EmiIngredientSerializer<T> {
 	static final Pattern STACK_REGEX = Pattern.compile("^([\\w_\\-./]+):([\\w_\\-.]+):([\\w_\\-./]+)(\\{.*\\})?$");
 	
-	EmiStack create(Identifier id, NbtCompound nbt, long amount);
+	EmiStack create(Identifier id, ComponentChanges componentPatch, long amount);
```
emilyploszaj: why componentPatch here

xplat/src/main/java/dev/emi/emi/platform/EmiAgnos.java
```diff
@@ -94,17 +96,17 @@ public static List<TooltipComponent> getItemTooltip(ItemStack stack) {
...
-	public static Text getFluidName(Fluid fluid, NbtCompound nbt) {
+	public static Text getFluidName(Fluid fluid, ComponentChanges nbt) {
 		return delegate.getFluidNameAgnos(fluid, nbt);
 	}
 
-	protected abstract Text getFluidNameAgnos(Fluid fluid, NbtCompound nbt);
+	protected abstract Text getFluidNameAgnos(Fluid fluid, ComponentChanges nbt);
 
-	public static List<Text> getFluidTooltip(Fluid fluid, NbtCompound nbt) {
+	public static List<Text> getFluidTooltip(Fluid fluid, ComponentChanges nbt) {
```
emilyploszaj: All these EmiAgnos methods still say `nbt`

emilyploszaj (commented):
Did another pass, there still are a handful of outstanding things I mentioned the first time unfixed.

xplat/src/main/java/dev/emi/emi/EmiPort.java
```diff
@@ -212,7 +218,20 @@ public static Comparison compareStrict() {
 		return Comparison.compareNbt();
 	}
 
+	public static ItemStack setPotion(ItemStack stack, RegistryEntry<Potion> potion) {
```
emilyploszaj: This extra method with registry entries shouldn't exist

xplat/src/main/java/dev/emi/emi/EmiPort.java
```diff
@@ -212,7 +218,20 @@ public static Comparison compareStrict() {
...
+		stack.apply(DataComponentTypes.POTION_CONTENTS, PotionContentsComponent.DEFAULT, potion, PotionContentsComponent::with);
+		return stack;
+	}
+
 	public static ItemStack setPotion(ItemStack stack, Potion potion) {
-		return PotionUtil.setPotion(stack, potion);
+		return setPotion(stack, getPotionRegistry().getEntry(potion));
+	}
+
+	public static @Nullable DynamicRegistryManager getRegistryManager() {
```
emilyploszaj: This new method shouldn't exist unless it's across all versions, which it can't be.

xplat/src/main/java/dev/emi/emi/EmiUtil.java
emilyploszaj: This is still present despite being in EmiPort

xplat/src/main/java/dev/emi/emi/api/stack/Comparison.java
```diff
@@ -8,7 +8,7 @@
 import dev.emi.emi.runtime.EmiLog;
 
 public class Comparison {
-	private static final Comparison COMPARE_NBT = Comparison.compareData(stack -> stack.getNbt()); 
+	private static final Comparison COMPARE_NBT = Comparison.compareData(stack -> stack.getComponentChanges());
```
emilyploszaj: This needs to be named `COMPARE_COMPONENTS` with the helper method to get it renamed as well.

xplat/src/main/java/dev/emi/emi/api/stack/EmiRegistryAdapater.java
```diff
@@ -25,7 +28,7 @@ public interface EmiRegistryAdapater<T> {
 	/**
 	 * Constructs an {@link EmiStack} from a given object from the registry, or {@link EmiStack#EMPTY} if somehow invalid.
 	 */
-	EmiStack of(T t, NbtCompound nbt, long amount);
+	EmiStack of(T t, @Nullable ComponentChanges componentChanges, long amount);
```
emilyploszaj: I think this shouldn't be nullable, I'll push a commit that has `EmiPort.emptyNbt()` which will return null in 1.20.4 and `ComponentChanges.EMPTY` in 1.20.5 which should reduce diff sizes (in additions to fixing the tag stack size issue).

emilyploszaj (changes requested):
This is pretty close, I've made some new comments for things that need to get fixed and will push a commit with some final abstractions like I mentioned with the dyes.

emilyploszaj (approved):

PR #519, opened 2024-04-20
Flatten ItemEmiStack and store item/NBT directly

embeddedt:
Reduces total retained size of `ItemEmiStack` objects in a test modpack from roughly 145MB -> 70MB. This change shouldn't be visible to mods unless they rely on consistent identity of the returned `ItemStack`.

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/api/stack/ItemEmiStack.java
+++ b/xplat/src/main/java/dev/emi/emi/api/stack/ItemEmiStack.java
@@ -44,10 +44,10 @@ public class ItemEmiStack extends EmiStack implements Batchable {
 	}
 
 	public ItemEmiStack(ItemStack stack, long amount) {
-		this(stack.getItem(), amount, stack.getNbt());
+		this(stack.getItem(), stack.getNbt(), amount);
 	}
 
-	public ItemEmiStack(Item item, long amount, NbtCompound nbt) {
+	public ItemEmiStack(Item item, NbtCompound nbt, long amount) {
 		this.item = item;
 		this.nbt = nbt != null ? nbt.copy() : null;
 		this.amount = amount;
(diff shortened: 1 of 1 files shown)


xplat/src/main/java/dev/emi/emi/api/stack/ItemEmiStack.java
```diff
@@ -31,29 +33,38 @@
...
 	}
 
 	public ItemEmiStack(ItemStack stack, long amount) {
-		stack = stack.copy();
-		stack.setCount((int) amount);
-		this.stack = stack;
+		this(stack.getItem(), amount, stack.getNbt());
+	}
+
+	public ItemEmiStack(Item item, long amount, NbtCompound nbt) {
```
emilyploszaj: I'd prefer `item, nbt, amount` for param order here.

xplat/src/main/java/dev/emi/emi/api/stack/ItemEmiStack.java
```diff
@@ -31,29 +33,38 @@
...
+		this.item = item;
+		this.nbt = nbt != null ? nbt.copy() : null;
 		this.amount = amount;
 	}
 
 	@Override
 	public ItemStack getItemStack() {
-		stack.setCount((int) amount);
+		ItemStack stack = new ItemStack(this.item, (int)this.amount);
```
emilyploszaj: I put a space after casts

emilyploszaj (approved):
Just some style nits. This was discussed elsewhere and while there will be some downsides I think it's reasonable. Will merge once comments are addressed.

emilyploszaj:
Er, sorry I just remembered there are EmiStack.of methods for items in EmiStack that wastefully construct an itemstack now that this new constructor exists, can you adjust those as well?

embeddedt:
I am not sure that adjustment would be safe as the default stack might contain NBT (which we cannot retrieve without constructing it, as far as I know).

emilyploszaj:
Hm. Yeah I guess that's fine then.

emilyploszaj:
I guess that means the new constructor should be private if we're assuming it can't be safely called

embeddedt:
I mean, it's fine to call if you have both the item and its NBT already, through some other means. It's no less safe than constructing a stack manually just to pass it to the constructor.

emilyploszaj:
...I just realized I'm talking about an EmiStack.of method that takes an item, nbt, and amount that I haven't committed yet. I'll handle it myself.

PR #536, opened 2024-05-12
Avoid constructing ItemStack just to check tag membership

embeddedt:
Since `getItemStack` is now rather complex (allocates an ItemStack), we may as well skip using it in spots where the stack is actually unnecessary.


emilyploszaj:
You say "spots" here, do you intend to update this PR with more than 1 change?

embeddedt:
I will make another PR if I do, because the other areas ended up requiring more effort to make the code clean (`EmiStack` needs to gain a new method returning the `Item` that `getItemStack` would return).

emilyploszaj:
It already exists, `EmiStack.getKey`. I'm not further exposing first class treatment of `EmiStack`s as glorified `ItemStack`s

emilyploszaj (approved):

PR #558, opened 2024-05-30
Port to 1.21

embeddedt:
This PR will track porting progress to 1.21. Everything compiles and (mostly) runs, however, there are quite a few bugs and things worth noting:

* Fabric's model loading API is not updated yet so the code for that is commented out.
* The `Identifier` constructor became private, which resulted in a lot of `new Identifier`->`Identifier.of` refactoring.
* Enchantments now need `RegistryEntry` to detect whether they are a curse, etc. At the moment I just retrieve the entry from the registry at each point it's needed rather than storing the `RegistryEntry`.
* The stack batcher is broken by the redesign of `VertexConsumerProvider.Immediate` in 24w21a; rendering the first batch works but it crashes as soon as it needs to update. Cause of this is not yet known. Rendering works fine with the batcher disabled, though.
* `EmiShapedRecipe.setRemainders` doesn't function and throws many errors because the new recipe input system resizes the inventory to fit the given recipe, which breaks the indexing logic.
* I have most likely changed a few things you'd rather be abstracted out into `EmiPort`. :)


xplat/src/main/java/dev/emi/emi/EmiPort.java
```diff
@@ -148,7 +148,7 @@ public static void setPositionTexShader() {
 	}
 
 	public static void setPositionColorTexShader() {
-		RenderSystem.setShader(GameRenderer::getPositionColorTexProgram);
```
emilyploszaj, line 151: Do any usages of this expect the arguments to be position color tex ordered?

emilyploszaj (changes requested):
I've refactored identifier construction in 1.20.5, left one comment, looks reasonable otherwise, thanks again for your contributions.

xplat/src/main/java/dev/emi/emi/EmiPort.java
embeddedt: Order doesn't matter as of 1.21 (which is why they got rid of this shader).

xplat/src/main/java/dev/emi/emi/EmiPort.java
```diff
@@ -168,7 +168,7 @@ public static Registry<Potion> getPotionRegistry() {
 	}
 
 	public static Registry<Enchantment> getEnchantmentRegistry() {
-		return Registries.ENCHANTMENT;
+		return MinecraftClient.getInstance().world.getRegistryManager().get(RegistryKeys.ENCHANTMENT);
```
Sollace, line 171: This does not seem to work, as I'm seeing it returning nulls for enchantment ids when loading a world.

embeddedt: It may be possible to obtain it off the connection instead; that might be available sooner.

emilyploszaj:
This PR is not going to be merged, as discussed, since you ended up being busy, I did the changes locally and will hopefully push soon. Thanks for your contribution though, I did use it to base my abstraction changes.

PR #595, opened 2024-06-28
zh_CN 1.1.8 update

IJustWantANamePlease:
Adds some missing and new strings & fixs some keys that I spelt wrong.

However, there is an issue which I need help on: some tags aren't getting translated:
 - tags "Mining Enchantable" and "Mining Looting Enchantable" translations aren't applied (probably why they were left untouched in the previous commit);
 - tag "Enchantables" isn't present in en_us.json, and uhh... I didn't manage to guess its key.


emilyploszaj:
I believe these tags are vanilla tags and are translated by neoforge and fabric. I will merge what you have however.

emilyploszaj (approved):

PR #624, opened 2024-07-10
Re-enable StackBatcher for 1.21

embeddedt:
This should get a second pass from someone familiar with 1.21 rendering to ensure there isn't an obvious issue, but it seems to work fine for vanilla content at least.

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/screen/StackBatcher.java
+++ b/xplat/src/main/java/dev/emi/emi/screen/StackBatcher.java
@@ -276,14 +276,14 @@ public class StackBatcher {
 		public VertexConsumer getBuffer(RenderLayer renderLayer) {
 			BufferBuilder bufferBuilder = this.pending.get(renderLayer);
 
-			if(bufferBuilder == null) {
+			if (bufferBuilder == null) {
 				BufferAllocator allocator = this.layerBuffers.get(renderLayer);
-				if(allocator != null) {
+				if (allocator != null) {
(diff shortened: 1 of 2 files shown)


xplat/src/main/java/dev/emi/emi/screen/StackBatcher.java
```diff
@@ -237,14 +214,18 @@ private void bake() {
 		for (Map.Entry<RenderLayer, BufferBuilder> entry : imm.getPendingLayerBuffers().entrySet()) {
 			bake(entry.getKey(), entry.getValue());
 		}
+		imm.getPendingLayerBuffers().clear();
 	}
 
 	public void bake(RenderLayer layer, BufferBuilder bldr) {
-		// TODO if (!imm.getActiveConsumers().remove(bldr)) return;
+		var builtBuffer = bldr.endNullable();
```
emilyploszaj: No var please

emilyploszaj (approved):
All good by me, just a single style nit, will merge when adjusted and do further testing.

xplat/src/main/java/dev/emi/emi/screen/StackBatcher.java
embeddedt: Fixed, but there are many other `var`s in the codebase by now. Admittedly, most of them seem related to 1.20.5 changes, which suggests I probably introduced them in the 1.20.5 port. :sweat_smile:

PR #630, opened 2024-07-16
Create pt_br.json for localization

Landrozaum:
Added pt_br.json for Brazillian Portuguese localization.


emilyploszaj:
This doesn't include translated config tooltips (see https://github.com/emilyploszaj/emi/blob/1.21/xplat/src/main/resources/assets/emi/lang/zh_cn.json for an example), is that intentional?

Landrozaum:
Sorry, I used en_us.json as a base and it doesn't have the tooltips, I'll fix that and add the lines.

Landrozaum:
Done, awaiting your approval.
😉

emilyploszaj (approved):

Issue #629, opened 2024-07-15
[1.1.10+1.21+fabric] Incompatible with Better Statistics Screen


luciobortoletto:
### What happened?

Minecraft crashes when opening inventory on Quilt (I haven't tested on Fabric). This doesn't happen when only one of the 2 mods (Better Statistics Screen or Emi) is installed.

### Steps to reproduce

1. Enter in a world
2. Press E to open the inventory
3. Minecraft crashes.

### Relevant log output or crash report

Log: https://mclo.gs/pRFCsFM
Crash report: https://mclo.gs/CZVRfwI

### Other installed mods

- Better Statistics Screen 3.12.3+fabric-1.21
- EMI 1.1.10+1.21+fabric
- Quilted Fabric API (QFAPI) / Quilt Standard Libraries (QSL) 11.0.0-alpha.3+0.100.7-1.21
- TCDCommons API 3.12.1+fabric-1.21  
  
I also opened the issue on Better Statistics Screen's repository: https://github.com/TheCSMods/mc-better-stats/issues/133

emilyploszaj:
This item group concurrency issue should be resolved in 1.1.11

Issue #620, opened 2024-07-06
Network Protocol Error


Wikinger8:
https://github.com/emilyploszaj/emi/assets/85295145/190baea7-db19-4bcc-85dc-78fc63d2643b

![image](https://github.com/emilyploszaj/emi/assets/85295145/812f63e5-9e0a-4e95-8723-040fda76e46e)
[latest.log](https://github.com/user-attachments/files/16117130/latest.log)
[disconnect-2024-07-06_21.33.51-client.txt](https://github.com/user-attachments/files/16117131/disconnect-2024-07-06_21.33.51-client.txt)

Somehow Sodium is the problem.

emilyploszaj:
Please report this to sodium.

Wikinger8:
![image](https://github.com/emilyploszaj/emi/assets/85295145/5bc88676-1106-4d3c-a68e-da41b17f2e42)
https://github.com/CaffeineMC/sodium-fabric/issues/2589

MeeniMc:
This can be replicated in singleplayer: the game will load to completion, and sometimes crash when first opening the inventory. 

Log contains the following error, which does not crash the game
```
[13:51:05] [Render thread/INFO] (EMI) [EMI] Tags synchronized, reloading EMI
[13:51:05] [Render thread/INFO] (EMI) [EMI] Starting Reload
[13:51:05] [Thread-17/INFO] (EMI) [EMI] Starting EMI reload...
[13:51:05] [Thread-17/INFO] (EMI) [EMI] Clearing data
[13:51:05] [Thread-17/INFO] (EMI) [EMI] Initializing plugin from emi
[13:51:05] [Thread-17/INFO] (EMI) [EMI] Initialized plugin from emi in 6ms
[13:51:05] [Thread-17/INFO] (EMI) [EMI] Initializing plugin from owo
[13:51:05] [Thread-17/INFO] (EMI) [EMI] Initialized plugin from owo in 0ms
[13:51:05] [Thread-17/INFO] (EMI) [EMI] Processing tags
[13:51:05] [Thread-17/INFO] (EMI) [EMI] Constructing index
[13:51:05] [Thread-17/ERROR] (EMI) [EMI] Creative item group Â§fColored Blocks threw while EMI was attempting to construct the index, items may be missing.
[13:51:05] [Thread-17/INFO] (Minecraft) [STDERR]: java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 513
[...]

PR #631, opened 2024-07-16
Update item group entries on client thread

embeddedt:
There seems to be an unwritten contract in the community that this method is only invoked on the main thread, and not doing that is causing several concurrency crashes (e.g. #629, #620) which usually get blamed on another mod. (Actually, the same concurrency issue would probably happen if the creative inventory screen were opened while EMI was still loading, unless EMI explicitly blocks this.)

As a stopgap, until some better solution is found, this PR moves the call to `updateEntries` to happen inside a task submitted to the main thread. This will probably introduce a couple seconds lag spike in a large enough modpack, but that is preferable to an undiagnosable crash.

Better solutions welcome.

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/registry/EmiStackList.java
+++ b/xplat/src/main/java/dev/emi/emi/registry/EmiStackList.java
@@ -8,6 +8,7 @@ import java.util.Set;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 
+import com.google.common.base.Stopwatch;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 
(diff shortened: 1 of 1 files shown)


emilyploszaj:
If this is going to happen with threading, groups should be queried after being updated, and if possible this should happen as early as possible and only joined when EMI is baking the groups. A bit more complex than your solution, but would have better performance for players (EMI can load while the groups are being built), and perhaps a bit concurrent safer (collecting contents right after creating them avoids EMI running into groups which are being updated on the main thread).

Edit: Just realized this is basically at the start of the reload, and since it is important to happen early, the threading solution present works. However, you should still call `ItemGroup.getSearchTabStacks()` immediately after updating and storing them in a map or something.

xplat/src/main/java/dev/emi/emi/registry/EmiStackList.java
```diff
@@ -76,14 +78,25 @@ public static void reload() {
...
+			// There is an unwritten convention that ItemGroup.updateEntries is only invoked on the main thread
+			Map<ItemGroup, Collection<ItemStack>> itemGroupToStacksMap = client.submit(() -> {
+				Map<ItemGroup, Collection<ItemStack>> map = new Reference2ReferenceOpenHashMap<>();
+				Stopwatch watch = Stopwatch.createStarted();
+				for (ItemGroup group : ItemGroups.getGroups()) {
+					group.updateEntries(context);
+					map.put(group, group.getSearchTabStacks());
+				}
+				watch.stop();
+				EmiLog.info("Reloading item group stacks took " + watch);
```
emilyploszaj: Logging is good, for consistency though, EMI has used `System.currentTimeMillis()` manual counting in all places so far. Definitely worse style but consistency is consistency, I'll consider a cleanup in the future.

Additionally, for steps like this that might take a while, I'd probably log "Reloading item groups" at the start and "Reloaded item groups in [time]ms" at the end (maybe outside of the thread so all logging happens in the EMI thread?)

emilyploszaj (approved):
One crit, otherwise good.

emilyploszaj (approved):

PR #646, opened 2024-08-01
Misc Jemi fixes

pupnewfster:
- Fixed an exception I noticed in my log when processing certain recipes via `JemiPlugin#addCraftingRecipes`, as since `RecipeEntry`'s have been a thing the recipes jei tracks are `RecipeEntry<CraftingRecipe>` and not just `CraftingRecipe`. (The unchecked cast hid this error at compile time)
- Fixed a breaking change to the internals of `SubtypeInterpreters` that made `contains` take the ingredient instead of the base element
- Accounted for a deprecation for removal since JEI 19.1.0 in the `JemiStackSerializer` code (as the fix to subtype interpreters needs a version of `19.3.0` or higher)
- Declared the new method in our implementation of `IRecipeGui` so that if anyone calls it they won't get an abstract method not implemented exception (I forget the exact type of exception). Though I am not exactly sure how to implement the actual logic for the method, so would appreciate guidance (for now I just put a TODO).

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/jemi/JemiStackSerializer.java
+++ b/xplat/src/main/java/dev/emi/emi/jemi/JemiStackSerializer.java
@@ -31,9 +31,9 @@ public class JemiStackSerializer implements EmiIngredientSerializer<JemiStack> {
 			if (type == VanillaTypes.ITEM_STACK || type == JemiUtil.getFluidType()) {
 				continue;
 			}
-			Optional<?> opt = manager.getIngredientByUid(type, uid);
+			Optional<EmiStack> opt = manager.getTypedIngredientByUid(type, uid).map(JemiUtil::getStack);
 			if (opt.isPresent()) {
-				return JemiUtil.getStack(type, opt.get()).setAmount(amount);
+				return opt.get().setAmount(amount);
 			}
 		}
 		return EmiStack.EMPTY;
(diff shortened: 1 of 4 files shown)


xplat/src/main/java/dev/emi/emi/jemi/JemiStackSerializer.java
```diff
@@ -31,9 +31,9 @@ public EmiStack create(String uid, long amount) {
 			if (type == VanillaTypes.ITEM_STACK || type == JemiUtil.getFluidType()) {
 				continue;
 			}
-			Optional<?> opt = manager.getIngredientByUid(type, uid);
+			Optional<EmiStack> opt = manager.getTypedIngredientByUid(type, uid).map(JemiUtil::getStack);
 			if (opt.isPresent()) {
-				return JemiUtil.getStack(type, opt.get()).setAmount(amount);
```
emilyploszaj, line 36: Is it reasonable to remove this explicit passing of the type in `JemiUtil.getStack`?

xplat/src/main/java/dev/emi/emi/jemi/runtime/JemiRecipesGui.java
```diff
@@ -45,6 +46,11 @@ public void showTypes(List<RecipeType<?>> recipeTypes) {
 		}
 	}
 
+	@Override
+	public <T> void showRecipes(IRecipeCategory<T> recipeCategory, List<T> recipes, List<IFocus<?>> focuses) {
+		//TODO: Figure out how to implement
```
emilyploszaj: Delegating this to just viewing the category is fine, EMI doesn't expose this type of granularity.

xplat/src/main/java/dev/emi/emi/jemi/JemiStackSerializer.java
pupnewfster: Yes, because `getTypedIngredientByUid` returns an optional of `ITypedIngredient`, so this ends up using the overload for `JemiUtil#getStack` that grabs the type from the ingredient

emilyploszaj (approved):

emilyploszaj:
Thank you for your contribution, I'll try to get a release of this out as soon as I have a proper chance.

PR #662, opened 2024-08-11
Continue processing item groups if one throws an error

embeddedt:
#631 introduced a regression that causes loading to fail completely if one item group throws an exception; this PR restores the `try`/`catch` around individual groups.

Duplication of the `groupName` logic is not ideal but I didn't think it was worth more refactoring to avoid it.


emilyploszaj (approved):
This will technically throw down the line if a group doesn't populate the map but it'll be caught, which I suppose is fine. I'll make further adjustments myself, thanks for the contribution.

PR #675, opened 2024-08-26
Fix typo in config tooltip: focuse -> focus

Mithi83:


emilyploszaj:
This was actually supposed to be "Focuses", so I adjusted it. Thanks for pointing it out.

Issue #709, opened 2024-09-21
[Crash :boom: ] EMI and Apothic Enchanting `LegacyRandomSource` crash


pietro-lopes:
Some interaction is going on there that is causing crash.

https://mclo.gs/00RRUFl (captured with BetterRandomSourceConcurrencyCrash mod)

Also, Apothic Enchanting dev said the following:
`"EMI should not be supplying the Level to the tooltip context if resolving it offthread"`

This crash is happening very frequently on ATM10

ghost:
Adding a log from a second user on the ATM discord that has this issue, it is just about identical to the already supplied one
https://mclo.gs/2QkUF8N

emilyploszaj:
Resolved in 1.1.14

Issue #575, opened 2024-06-17
ToolTip Context Not Passing Context


fzzyhmstrs:
Checking on the log below from Technici4n, it looks like `ItemEmiStack` (and maybe elsewhere? Didn't check) uses `TooltipContext.DEFAULT` in it's tooltip method. This short circuits any possibility of context in the context.

`TooltipContext.create(client.world)`, which takes a nullable param, would build contextual Contexts as needed by users.

https://github.com/emilyploszaj/emi/blob/615d816a81de7627292934c1c3d99b4b6931cfd1/xplat/src/main/java/dev/emi/emi/api/stack/ItemEmiStack.java#L162

```
java.lang.NullPointerException: Cannot invoke "net.minecraft.core.HolderLookup$Provider.lookupOrThrow(net.minecraft.resources.ResourceKey)" because the return value of "net.minecraft.world.item.Item$TooltipContext.registries()" is null
    at TRANSFORMER/modern_industrialization@0.0.0-local/aztech.modern_industrialization.items.diesel_tools.DieselToolItem.appendHoverText(DieselToolItem.java:134)
    at TRANSFORMER/minecraft@1.21/net.minecraft.world.item.ItemStack.getTooltipLines(ItemStack.java:772)
    at TRANSFORMER/emi@1.1.7+1.21+neoforge/dev.emi.emi.api.stack.ItemEmiStack.getTooltipText(ItemEmiStack.java:162)
    at TRANSFORMER/emi@1.1.7+1.21+neoforge/dev.emi.emi.search.EmiSearch.bake(EmiSearch.java:62)
    at TRANSFORMER/emi@1.1.7+1.21+neoforge/dev.emi.emi.runtime.EmiReloadManager$ReloadWorker.run(EmiReloadManager.java:222)
    at java.base/java.lang.Thread.run(Thread.java:1583)
    ```

emilyploszaj:
This was changed to provide context if possible in 1.1.8

PR #710, opened 2024-09-22
Do not provide world context during off-thread tooltip retrieval

embeddedt:
The world/entity are not thread safe, so they should not be provided to `Item#getTooltip` calls from the search thread.

Note that the issue described in #709 mainly affects NeoForge, as it exposes the passed world to mods, while vanilla just uses it for the tick rate & registry managers. https://github.com/neoforged/NeoForge/blob/42023dc3e3284b7046791403db04649999389ba3/patches/net/minecraft/world/item/Item.java.patch#L221-L222

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/api/stack/ItemEmiStack.java
+++ b/xplat/src/main/java/dev/emi/emi/api/stack/ItemEmiStack.java
@@ -159,7 +159,12 @@ public class ItemEmiStack extends EmiStack implements Batchable {
 
 	@Override
 	public List<Text> getTooltipText() {
-		return getItemStack().getTooltip(Item.TooltipContext.create(client.world), client.player, TooltipType.BASIC);
+		if (client.isOnThread()) {
+			return getItemStack().getTooltip(Item.TooltipContext.create(client.world), client.player, TooltipType.BASIC);
+		} else {
+			// Don't provide world or entity as context, as they are not thread safe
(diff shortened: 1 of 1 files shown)


emilyploszaj:
Is this an acceptable resolution? If I remember correctly, this context was added in the past due to errant mods crashing without it, will this not just revert that behavior? This seems like a rock and a hard place of trusting downstream mods and needing multithreading. Insight would be appreciated.

SiverDX:
But doesn't vanilla do this as well (on 1.21 at least)?
`SessionSearchTrees -> updateRecipes / updateCreativeTooltips`

NeoForge has annotated the method itself as `@Nullable` as well so personally I'd say it falls on the mods to handle this properly
Also as I understand it this is mostly an 1.21 issue since the level is provided by NeoForge (which means if mods have problems it would be reasonable to expect them to fix this themselves since it is the latest version as of now)

SiverDX:
You seem to have added the level for https://github.com/emilyploszaj/emi/issues/575
But the PR would use vanilla behaviour which still supplies access to the registries

embeddedt:
As above, we still provide registry context (as the registry should be thread-safe) but we cannot provide the whole level safely when not on the main thread, as mods can easily do things like roll the RNG in it.

emilyploszaj:
Alright, reasonable enough. I'll be getting crash reports of one of the two either way.

emilyploszaj (approved):

PR #731, opened 2024-10-03
Add dev mode nag for recipe IDs that should be synthetic

### LINKED ISSUE #730: Nags for recipe IDs that should be synthetic
As per the EmiRecipe javadoc
https://github.com/emilyploszaj/emi/blob/6127e8f6a9b736140c384f58d20faadd61add2db/xplat/src/main/java/dev/emi/emi/api/recipe/EmiRecipe.java#L23

Check that every registered EmiRecipe with a non-synthetic ID actually exists by looking it up in the recipe manager. Log those that don't.

Abbie5:
Closes #730

Also changes some brewing recipe IDs to be synthetic.

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/registry/EmiRecipes.java
+++ b/xplat/src/main/java/dev/emi/emi/registry/EmiRecipes.java
@@ -156,6 +156,7 @@ public class EmiRecipes {
 			this.recipes = List.copyOf(recipes);
 	
 			Object2IntMap<Identifier> duplicateIds = new Object2IntOpenHashMap<>();
+			Set<Identifier> missingIds = new ObjectArraySet<>();
 			for (EmiRecipe recipe : recipes) {
 				Identifier id = recipe.getId();
 				EmiRecipeCategory category = recipe.getCategory();
@@ -179,8 +180,8 @@ public class EmiRecipes {
 						byId.put(id, recipe);
 					}
 
-					if (EmiConfig.devMode && !id.getPath().startsWith("/") && !recipeIds.containsValue(id)) {
-						EmiReloadLog.warn("Recipe " + id + " not present in recipe manager. Consider prefixing its path with '/' to signify that it is synthetic.");
+					if (!id.getPath().startsWith("/") && !recipeIds.containsValue(id)) {
+						missingIds.add(id);
 					}
 				}
 			}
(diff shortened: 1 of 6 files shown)


xplat/src/main/java/dev/emi/emi/screen/WidgetGroup.java
```diff
@@ -79,6 +79,10 @@ public void decorateDevMode() {
 				tooltip.addAll(EmiTooltip.splitTranslate("emi.dev.synthetic_id"));
 			}
 			errors.add(new RecipeError(RecipeError.Severity.ERROR, tooltip));
+		} else if (EmiDev.missingRecipeIds.contains(id)) {
+			List<TooltipComponent> tooltip = Lists.newArrayList();
+			tooltip.addAll(EmiTooltip.splitTranslate("emi.dev.synthetic_id"));
```
emilyploszaj, line 85: Please add a more detailed tooltip for what's going on and why a dev should use a synthetic ID.

xplat/src/main/java/dev/emi/emi/runtime/dev/EmiDev.java
```diff
@@ -6,4 +6,5 @@
 
 public class EmiDev {
 	public static Set<Identifier> duplicateRecipeIds = Set.of();
+	public static Set<Identifier> missingRecipeIds = Set.of();
```
emilyploszaj: All instances of `missingRecipeIds` and similar should probably be renamed, they're not missing IDs (i.e. null). Can't think of a great name, but `incorrectRecipeIds` would be better, if you have something you think suits it better go for it.

xplat/src/main/java/dev/emi/emi/registry/EmiRecipes.java
```diff
@@ -178,13 +179,20 @@ public Manager(List<EmiRecipeCategory> categories, Map<EmiRecipeCategory, List<E
...
+					}
 				}
 			}
 	
 			if (EmiConfig.devMode) {
 				for (Identifier id : duplicateIds.keySet()) {
 					EmiReloadLog.warn(duplicateIds.getInt(id) + " recipes loaded with the same id: " + id);
 				}
+				for (Identifier id : missingIds) {
+					EmiReloadLog.warn("Recipe " + id + " not present in recipe manager. Consider prefixing its path with '/' to signify that it is synthetic.");
```
emilyploszaj: Nit: perhaps change to "Consider prefixing its path with '/' if it is synthetic"

emilyploszaj (changes requested):
Thanks for the contribution and for finding all of the non-synthetic IDs remaining in EMI. Few crits then I'll merge.

Abbie5:
Sorry for taking so long, I've made the requested changes now.

emilyploszaj (approved):

PR #750, opened 2024-10-12
Search Box Suggestion Display Fix (1.20)

AugustSaintFreytag:
This pull request fixes the placeholder/suggestion text (e.g. "Search EMI…") being visible together with the insertion point, overlapping each other. These changes make it so the `TextFieldWidget` used for EMI's search now hides its suggestion text whenever the field has focus, when the player wants to type. Once it loses focus, the suggestion is displayed again. Existing logic for hiding the suggestion when the text field has input is kept as-is.

I've also prepared this in basically the same way for `1.21` if you want me to open the pull request myself, though these changes are minor and you might be faster in your own process adding them.


emilyploszaj:
Please open PRs targeting the latest version.

PR #752, opened 2024-10-13
Search Box Suggestion Display Fix (1.21)

AugustSaintFreytag:
This pull request fixes the placeholder/suggestion text (e.g. "Search EMI…") being visible together with the insertion point, overlapping each other. These changes make it so the `TextFieldWidget` used for EMI's search now hides its suggestion text whenever the field has focus, when the player wants to type. Once it loses focus, the suggestion is displayed again. Existing logic for hiding the suggestion when the text field has input is kept as-is.

I’m opening this because the one for 1.20 was rejected and closed.

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/screen/widget/EmiSearchWidget.java
+++ b/xplat/src/main/java/dev/emi/emi/screen/widget/EmiSearchWidget.java
@@ -81,15 +81,20 @@ public class EmiSearchWidget extends TextFieldWidget {
 		});
 		this.setChangedListener(string -> {
 			if (string.isEmpty()) {
-				this.setSuggestion(I18n.translate("emi.search"));
+				if (!this.isFocused()) {
+					this.setSuggestion(I18n.translate("emi.search"));
+				}
+				
 				EmiScreenManager.focusSearchSidebarType(EmiConfig.emptySearchSidebarFocus);
 			} else {
 				this.setSuggestion("");
 				EmiScreenManager.focusSearchSidebarType(EmiConfig.searchSidebarFocus);
 			}
+
 			Matcher matcher = EmiSearch.TOKENS.matcher(string);
 			List<Pair<Integer, Style>> styles = Lists.newArrayList();
 			int last = 0;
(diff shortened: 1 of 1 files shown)


AugustSaintFreytag:
Apparently, the same change would have to be done for the config field, as well, it has the same issue.

emilyploszaj:
The title calls this change a "fix" but I'm not sure either behavior is more correct. Does vanilla have examples where it might disagree with this, or is there something else justifying this? It is probably just me being used to it, but I kinda prefer it as is.

AugustSaintFreytag:
Well, the insertion point overlaps with the placeholder text, looks like a visual bug to me. I've got that from other players of the mod pack I'm maintaining right now. Generally, in input fields where the placeholder text stays, the insertion point is a vertical bar, not an underscore, so there it can't intersect. Vanilla Minecraft doesn't have a lot of examples, it doesn't use many text fields and some don't have a placeholder at all (like chat input).

I took the effort to go through gameplay recordings to give you the most direct comparison, this is REI:

ezgif-2-d20707ce19

Their devs seem to have either used an existing option to use a vertical bar as an insertion point or actually built their own little rendering logic for it.

From a UX standpoint, it's simple. If the search box is not focussed, it may be useful to convey what it does with the placeholder label. If the player clicks the search box, they want to type. They already know what the search box is for, they don't need a placeholder.

SiverDX:
Don't have a personal opinion on this but in the world creation screen in 1.21 (maybe 1.20 too? not sure) it looks like this:

image

image

emilyploszaj:
Alright that's reasonable to me then, I will request you remove the non-functionality formatting changes made to the file

AugustSaintFreytag:
Am I going to have to write the changes for the config text field, too? And then open eight individual pull requests for every version of Minecraft EMI supports?

emilyploszaj:
I handle all the versioning, if it gets merged into 1.21 then it'll get merged backwards. Adjusting the config text field is your choice, I just asked for the formatting changes to be removed.

PR #758, opened 2024-10-18
Update and clean up `zh_cn.json`

sendevia:


emilyploszaj:
Why does this translation PR remove a ton of tooltips?

sendevia:
Because I noticed that there were no corresponding entries in `en_us.json`, but I have restored them now, sorry for my carelessness. In the latest commit I also optimized the use of punctuation/spacing a bit, it seems to be fine now.

emilyploszaj (approved):
Looks reasonable then. Tooltip translations are only present in non-english languages because they're handled by the default english config comment system otherwise.

PR #788, opened 2024-11-18
Add ability to disable tooltip indexing

embeddedt:
Reading the tooltips of modded items can be slow since many mods have relatively inefficient tooltip handlers. This PR adds an option to disable that, at the cost of preventing tooltips from being searched at all. The default is to still index tooltips.


emilyploszaj:
Notable perf improvement, but I think this is maybe an antifeature. Instead of having users disable features for better performance, perhaps search baking should be off of the reload thread like recipe baking falling back to manual matching (or simply not allowing search)? I can look into an implementation of this, I think I'd probably want my hands on it. What do you think?

PR #792, opened 2024-11-21
Add ingredient serialiser for list ingredients

Abbie5:


emilyploszaj:
This looks really reasonable, I'm wondering if there's any benefit to not serializing list ingredients? For example, EMI list ingredients are basically a fallback, and it'd be preferable if resource driven recipes and such only could add tag ingredients. Also, shouldn't this attempt to coerce ingredient lists to tags? Maybe lists of stacks for ingredient accepting values should be natively supported and coerced through several heuristics?

Abbie5:
Hmm, I was operating on the assumption that deserialising then reserialising should ideally return the input (although to do that properly would require making `ListEmiIngredient#ingredients` readable somehow, probably through a getter). This behaviour can be changed easily though, just by using `EmiIngredient::of` in place of `new ListEmiIngredient`

IMO, list ingredients should be able to be defined in resource-driven recipes, as tags cannot be added client-side.

I was a bit afraid of accidentally changing existing behaviour by interpreting lists of stacks as list ingredients, so I left it out of the PR. It can be added, though.

Abbie5:
implemented the changes i mentioned, sorry for the long delay. tested in both world interaction and info recipes, as well as favouriting list ingredients.

emilyploszaj (approved):

emilyploszaj:
Thanks for this, sorry for the long delay on my end

PR #812, opened 2024-12-23
Fix tag ingredient rendering in Fabulous

### LINKED ISSUE #811: EMI won't display tag items on recipes on fabulous graphics
tested on 1.20.1, both (neo)forge and fabric

With fabulous graphics enabled, some recipes that uses tags will not show the items in it, the items are invisible.

2D85FA0A-7778-4BDC-8BCB-7E816EA66DFA

image

image

embeddedt:
Fixes #811 

Based on comparisons with vanilla's item rendering logic, the wrong render layer was being used, which I suspect caused the items to get drawn on the item entity framebuffer, rather than the main framebuffer.


emilyploszaj (approved):
Oops, thanks for catching this.

PR #813, opened 2024-12-23
Populate item groups in correct order

### LINKED ISSUE #774: EMI appears to not be resetting cache on leaving worlds, which causes problems on relog
### Describe the bug, attach relevant log files, and describe what environment the issue occurs in

image

This is what the item component is supposed to look like. It's just a `Holder<Gene>`, where Gene is a data pack registry entry.

image

If I quit the world and rejoin (or rather, if I join *any* world/server after quitting one), it looks like this instead.

If I try to spawn this in, I get kicked with [this disconnect log](https://gist.github.com/aaronhowser1/7ee8fedfcf522b41a5489e8a400815d9). I also can't see any recipes that use that item. Note that this only happens to the items in EMI. Items in the creative menu or inventory work fine, and the recipes still work.

However, if I run `/reload` or `F3+T`, it goes back to working fine, looking like the first image. I can spawn it in and see it in recipes again.

I asked a friend of mine to check their in their own dev environment for their mod, and they also had the same issue, they just hadn't noticed due to how obscure the problem is.

embeddedt:
Since vanilla's search `ItemGroup` pulls data from the other item groups, we have to be careful to update the groups' entries in the correct order, otherwise the search group will persist outdated stacks from the previous reload. This change fixes #774.

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/registry/EmiStackList.java
+++ b/xplat/src/main/java/dev/emi/emi/registry/EmiStackList.java
@@ -94,7 +94,7 @@ public class EmiStackList {
 						EmiLog.error(e);
 					}
 				};
-				var itemGroups = ItemGroups.getGroups();
+				List<ItemGroup> itemGroups = ItemGroups.getGroups();
 				// Category item groups must be updated before non-category ones, otherwise the search group will
 				// read outdated item lists
 				itemGroups.stream().filter(g -> g.getType() == ItemGroup.Type.CATEGORY).forEach(itemGroupConsumer);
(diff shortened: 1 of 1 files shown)


xplat/src/main/java/dev/emi/emi/registry/EmiStackList.java
```diff
@@ -92,7 +93,12 @@ public static void reload() {
 						EmiLog.error("Creative item group " + groupName + " threw while EMI was attempting to construct the index, items may be missing.");
 						EmiLog.error(e);
 					}
-				}
+				};
+				var itemGroups = ItemGroups.getGroups();
```
emilyploszaj: EMI codebase doesn't use `var`.

emilyploszaj (commented):
Very peculiar design choice, thanks for figuring this out.

xplat/src/main/java/dev/emi/emi/registry/EmiStackList.java
embeddedt: Sorry, I intended to remove it before committing, and forgot. :sweat_smile:

emilyploszaj (approved):

PR #816, opened 2025-01-03
Add exclusion areas for toasts

Abbie5:
In my opinion, this is preferable to just hiding them entirely when EMI is visible.


emilyploszaj:
I think I disagree, toasts, especially recipe toasts, happen very frequently and often and would cause the UI to move significantly. As a correction, toasts are *delayed* instead of just hidden, they pop up when you close your inventory. I'm currently satisfied with this solution unless there's some problem I'm not taking into consideration. I'll close this, sorry :(

PR #839, opened 2025-01-25
Only store incorrect IDs in dev mode

embeddedt:
The `incorrectIds` set was being populated despite only being used when `EmiConfig.devMode` is enabled. This took a noticeable amount of time in a modpack.

image


emilyploszaj (approved):
Surprising cost, thanks for profiling

PR #854, opened 2025-02-07
Add top search bar config

yuko1101:
Currently there is center search bar option in the config, but we can't have it on the top of the screen i guess.

Most localizations are done by machine translator, so need to be checked.


emilyploszaj:
There should probably be adjustments to sidebar positioning rather than relying on automatic nudging from the search bar, additionally, EMI doesn't use machine translation

PR #863, opened 2025-02-15
Improve time complexity of EmiRegistry.addEmiStackAfter

embeddedt:
At the time of calling this, `EmiStackList.stacks` is still a linked list, so the time complexity of `get` scales with the provided index, resulting in the original loop being O(n^2). The new loop uses `ListIterator`, which allows traversing the linked list a single time and inserting directly after the desired node, rather than calling `List.add` which then requires another traversal through the list.

This improves the load time of Mahou Tsukai's EMI plugin from 19 seconds (before this change) to 60 milliseconds. However, the API should still be avoided in most cases, as its runtime cannot easily be improved beyond O(n).

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/registry/EmiRegistryImpl.java
+++ b/xplat/src/main/java/dev/emi/emi/registry/EmiRegistryImpl.java
@@ -1,5 +1,6 @@
 package dev.emi.emi.registry;
 
+import java.util.ListIterator;
 import java.util.function.Consumer;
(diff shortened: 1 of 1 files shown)


emilyploszaj (approved):
A proper reimplementation to store these additions in a map and adding them in a single pass might be in order. Would involve tossing errors down the line. This is a good perf improvement bandaid for the time being though, no clue how I let myself index a linked list.

PR #907, opened 2025-03-28
Silently drop packets going to non-EMI clients

### LINKED ISSUE #889: Not possible to join server with EMI on a client without it
Environment: 1.21.1, NeoForge 21.1.132
Mods: EMI 1.1.20+1.21.1

Server log: [latest.log](https://github.com/user-attachments/files/19145859/latest.log)
Client log: [latest.log](https://github.com/user-attachments/files/19145864/latest.log)

embeddedt:
This fixes #889 as the ping packet is no longer sent to a client without that channel registered. Design-wise, I'm unsure if it's better to put this check at the call sites of `EmiNetwork.sendToClient` or to do it here where we actually send the packet. I opted for the current approach since it means a graceful fallback will occur for any packet, not just the ones the check is explicitly added to.


Kowal198389 (approved):

emilyploszaj:
I actually implemented this locally a couple days ago, I'll commit it for the next release which hopefully will be soon, thanks though :)

PR #918, opened 2025-04-05
Fix stacks with a max size of 1 not being filled

PrototypeTrousers:
After the check damageable tools, like GT tools, get filled correctly


emilyploszaj:
Thanks for looking into this! I've already applied this change locally, so I'll close this PR, this functionality will come in the next update soon.

PR #928, opened 2025-04-23
add ability to view tree amounts in stack form

Vap0r1ze:
looks like this when alt is held:
ShareX_LfQZHD2uNY


SergioJuniorCE:
Is this still being considered?

PR #933, opened 2025-05-02
Deduplicate tag ingredient EmiStack lists

embeddedt:
This will likely need some refinement before being merged; opening it now to collect feedback.

This PR implements deduplication of the underlying stack lists used by `TagEmiIngredient` after all recipes are constructed. Initial testing suggests it saves about 80MB of memory in Craftoria; gains should scale with packs containing larger tags that are used often. The `TagEmiIngredient` objects themselves are still duplicated, but their shallow size isn't the majority of memory usage from what I see.

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/api/stack/TagEmiIngredient.java
+++ b/xplat/src/main/java/dev/emi/emi/api/stack/TagEmiIngredient.java
@@ -79,6 +79,11 @@ public class TagEmiIngredient implements EmiIngredient {
 		return stacks;
 	}
 
+	@ApiStatus.Internal
+	public void setEmiStacks(List<EmiStack> stacks) {
+		this.stacks = stacks;
+	}
+
 	@Override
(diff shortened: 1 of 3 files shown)


emilyploszaj:
Seems like a reasonable performance target, hashing the lists seems perhaps a bit wasteful though, surely ingredients could do lookups via tag key and then do comparison, or perhaps ignore all ingredients with mutated (chanced, remainder) elements

embeddedt:
I tried this and it's unfortunately a weaker optimization. Hashing the lists is more powerful as it can correctly deduplicate cases where many tag ingredients have mutated the ingredients in the exact same way (e.g. in Monifactory, there are a bunch of wrench tag ingredients where all the inner stacks have remainders with the wrench damaged by one more unit). The lookup approach barely saves anything in Monifactory, by comparison.

SlayerTheChikken:
o.o

PR #935, opened 2025-05-03
fix invalid "invalid ID" errors

screret:
Java's built in `IdentityHashMap` compares both keys *and values* by identity, which breaks mods that recreate the recipe ID for their EMI displays (it logs an invalid id error for every single recipe in GT.)
I fixed it by using a FastUtil `Reference2ObjectOpenHashMap` instead, which compares keys by reference and values with `equals()`.

This might slow the recipe loading down a bit, I'm not quite sure. If it's too slow, I'll implement something different on my end.

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/registry/EmiRecipes.java
+++ b/xplat/src/main/java/dev/emi/emi/registry/EmiRecipes.java
@@ -36,6 +36,7 @@ import it.unimi.dsi.fastutil.objects.Object2IntMap;
 import it.unimi.dsi.fastutil.objects.Object2IntOpenHashMap;
 import it.unimi.dsi.fastutil.objects.Object2ObjectOpenCustomHashMap;
 import it.unimi.dsi.fastutil.objects.ObjectArraySet;
+import it.unimi.dsi.fastutil.objects.Reference2ObjectOpenHashMap;
 import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.resource.language.I18n;
 import net.minecraft.recipe.Recipe;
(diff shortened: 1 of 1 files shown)


embeddedt:
I'm not understanding exactly what this PR is fixing. Why does it matter how `IdentityHashMap` compares values, when the primary use of this map is to retrieve recipe IDs via the `RecipeEntry` key (which should definitely be safe to compare via identity, as far as I know)?

emilyploszaj:
Agree with the above, as far as I know, this map only exists so that EMI can have a version agnostic way of getting a recipe's ID in the vanilla plugin, I'm going to close this PR.

fzzyhmstrs:
https://github.com/emilyploszaj/emi/blob/2fdde6a7461b4be62b374e68499291887f400dca/xplat/src/main/java/dev/emi/emi/registry/EmiRecipes.java#L184

It checks the value in the map right here. If using an `IdentityHashMap`, and the `EmiRecipe` builds a new id instead of referencing the recipe entries, the warning will be thrown even if the two ids are equal via `equals`

embeddedt:
Ah, yeah, in dev mode that's a valid issue, and the fix in the PR looks correct.

emilyploszaj:
I swear those were different maps... guess it got "optimized"

screret:
thank you! and sorry I couldn't answer the (now obvious) question of "why??"
I forgot to make the PR when I tested it works, remembered at 12am and immediately went to sleep after.
I guess writing a good PR description isn't the easiest thing to do at 12am.

PR #938, opened 2025-05-04
Nonexistent recipes with enchanted books in anvil menu

Sleepy-Horse:
Fix a bug where combining an item with an enchanted book in the anvil menu generated nonexistent recipes in survival mode.

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/platform/EmiAgnos.java
+++ b/xplat/src/main/java/dev/emi/emi/platform/EmiAgnos.java
@@ -159,7 +159,5 @@ public abstract class EmiAgnos {
 		return delegate.isEnchantableAgnos(stack, enchantment);
 	}
 
-	protected boolean isEnchantableAgnos(ItemStack stack, Enchantment enchantment) {
-		return enchantment.isAcceptableItem(stack) && stack.isEnchantable() && stack.getItem().isEnchantable(stack);
-	}
+	protected abstract boolean isEnchantableAgnos(ItemStack stack, Enchantment enchantment);
 }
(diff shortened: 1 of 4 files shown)


xplat/src/main/java/dev/emi/emi/platform/EmiAgnos.java
```diff
@@ -155,4 +154,12 @@ public static BakedModel getBakedTagModel(Identifier id) {
 	}
 
 	protected abstract BakedModel getBakedTagModelAgnos(Identifier id);
+
+	public static boolean isEnchantable(ItemStack stack, Enchantment enchantment) {
+		return delegate.isEnchantableAgnos(stack, enchantment);
+	}
+
+	protected boolean isEnchantableAgnos(ItemStack stack, Enchantment enchantment) {
```
embeddedt: Doesn't this also need a Fabric override?

Sleepy-Horse: I couldn't find any Fabric specific methods related to item enchantability using enchanted books in an anvil recipe. I only found the vanilla ones, which are unrelated to books. For instance, [this mod](https://github.com/MehVahdJukaar/Supplementaries/blob/ca051d9460465b0a6977c7e1fec5880593eb2a5e/common/src/main/java/net/mehvahdjukaar/supplementaries/common/items/KeyItem.java#L33C1-L37C6) implements this book logic only for Forge / Neoforge (`@ForgeOverride`). We may consider `bookEnchantable` to be always true unless an override is implemented.

emilyploszaj:
As a style convention agnos methods should be abstract and fully implemented in all cases. Additionally, any common functionality should not exist within agnos methods.

emilyploszaj (approved):
Thanks for this change

PR #946, opened 2025-05-12
Better support for opening EMI screens from non-handled screens

Abbie5:
This fixes the behaviour where, upon closing the recipe screen that was opened from the FTB Quests screen, the client is returned to the inventory screen instead of the original quest screen.

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/api/EmiApi.java
+++ b/xplat/src/main/java/dev/emi/emi/api/EmiApi.java
@@ -107,10 +107,10 @@ public class EmiApi {
 		Screen s = client.currentScreen;
 		if (s instanceof HandledScreen<?> hs) {
 			return hs;
-		} else if (s instanceof RecipeScreen rs) {
-			return rs.old;
-		} else if (s instanceof BoMScreen bs) {
-			return bs.old;
+		} else if (s instanceof RecipeScreen rs && rs.old instanceof HandledScreen<?> hs) {
+			return hs;
+		} else if (s instanceof BoMScreen bs && bs.old instanceof HandledScreen<?> hs) {
+			return hs;
 		}
 		return null;
 	}
@@ -243,13 +243,7 @@ public class EmiApi {
 			.collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue()));
 		if (!recipes.isEmpty()) {
 			EmiSidebars.lookup(stack);
-			if (getHandledScreen() == null) {
-				client.setScreen(new InventoryScreen(client.player));
-			}
-			if (client.currentScreen instanceof HandledScreen<?> hs) {
-				push();
-				client.setScreen(new RecipeScreen(hs, recipes));
-			} else if (client.currentScreen instanceof BoMScreen bs) {
+			if (client.currentScreen instanceof BoMScreen bs) {
 				push();
 				client.setScreen(new RecipeScreen(bs.old, recipes));
 			} else if (client.currentScreen instanceof RecipeScreen rs) {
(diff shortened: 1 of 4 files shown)


xplat/src/main/java/dev/emi/emi/registry/EmiRecipeFiller.java
```diff
@@ -97,6 +97,7 @@ public static <T extends ScreenHandler> List<EmiRecipeHandler<T>> getAllHandlers
 
 	@SuppressWarnings("unchecked")
 	public static <T extends ScreenHandler> @Nullable EmiRecipeHandler<T> getFirstValidHandler(EmiRecipe recipe, HandledScreen<T> screen) {
+		if (screen == null) return null;
```
emilyploszaj: Nit: EMI codebase style has brackets for all if statements

emilyploszaj (changes requested):
This likely works to fix this issue, but does it cause other issues? for instance, if you open from another GUI, it'll show no recipes as craftable, even if they may be craftable via the currently open screen handler? maybe getHandledScreen should default to that so we can support both use cases.

Abbie5:
Hi, coming back to this PR after a while so I'm a bit rusty. 

> This likely works to fix this issue, but does it cause other issues? for instance, if you open from another GUI, it'll show no recipes as craftable, even if they may be craftable via the currently open screen handler? maybe getHandledScreen should default to that so we can support both use cases.

Is there a way to test this behaviour so I can make sure my implementation is correct? And, I don't see a way to get a HandledScreen from the currently open screen handler exposed in PlayerEntity#currentScreenHandler, so please clarify what you're referring to by "that".

xplat/src/main/java/dev/emi/emi/api/EmiApi.java
```diff
@@ -107,10 +107,10 @@ public static HandledScreen<?> getHandledScreen() {
 		Screen s = client.currentScreen;
 		if (s instanceof HandledScreen<?> hs) {
 			return hs;
-		} else if (s instanceof RecipeScreen rs) {
-			return rs.old;
-		} else if (s instanceof BoMScreen bs) {
-			return bs.old;
+		} else if (s instanceof RecipeScreen rs && rs.old instanceof HandledScreen<?> hs) {
```
emilyploszaj, line 110: Here's what I mean in my previous comment, both RecipeScreen and BoMScreen can now have an old screen that is not a HandledScreen, such as a quest UI, and then no recipes would show up as craftable. Maybe this behavior should default to the player's inventory when nothing else is valid, since that is always accessible. WDYT?

Abbie5: like this?
```patch
Index: xplat/src/main/java/dev/emi/emi/api/EmiApi.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/xplat/src/main/java/dev/emi/emi/api/EmiApi.java b/xplat/src/main/java/dev/emi/emi/api/EmiApi.java
--- a/xplat/src/main/java/dev/emi/emi/api/EmiApi.java	(revision 12e249c067862fd4233bfb86ca8c92865f52fc8e)
+++ b/xplat/src/main/java/dev/emi/emi/api/EmiApi.java	(date 1771852166073)
@@ -111,6 +111,8 @@
 			return hs;
 		} else if (s instanceof BoMScreen bs && bs.old instanceof HandledScreen<?> hs) {
 			return hs;
+		} else if (client.player != null) {
+			return new InventoryScreen(client.player);
 		}
 		return null;
 	}
```

Abbie5: I just realised, mods can open the recipe UI from anywhere with ie. EmiApi::displayRecipe, personally I would find it disorienting if closing the recipe UI sent me to the inventory when I didn't have that open before.

PR #951, opened 2025-05-15
Do not retain JEMI layout builder for every recipe

embeddedt:
The `builder` field of `JemiRecipe` is only ever referenced during the constructor, and appears to just be used to compute the inputs/outputs/catalysts of the recipe. To actually render the recipe a new builder is created in `addWidgets`. Therefore it seems there's no need to retain this temporary initial builder in a field forever. Removing it substantially reduces the retained size of `JemiRecipe`s in a modpack.


emilyploszaj:
Well this seems like a really funny oversight, thanks for finding it.

PR #959, opened 2025-05-17
🔖 Search Bookmarks

rubybrowncoat:
This adds a new Panel/Sidebar that allows users to save their search queries for quick reuse.

This for example is how I'm currently using it:
image


emilyploszaj:
Do not call this feature "bookmarks", that's already a nickname for favorites, there is no reason this should be exposed to the API, please provide a better visualization method than a cycling ingredient, please adjust your type to not crash by returning null for no reason.

Additionally, is this a good solution for problems? EMI already saves recent searches and can cycle through them with up and down on the keyboard, desired stacks can be favorited and sorted manually, and the UI is optimized for this. Is there actually a reasonable use case that would justify a player adding a new sidebar just for search history? Is this even worth the implementation and maintenance burden?

ICaxapI:
In defense of this functionality, I can say that in technical builds, it's very convenient to save favorite bookmarks for certain complex searches, such as those related to processing different ores, to quickly switch between them when needed. For example, I could forget what variants were available for the copper I processed six months ago and simply apply the search I set up back then with one click.
By the way, it would be convenient if this search processed regular expressions.

Mickeon:
> EMI already saves recent searches and can cycle through them with up and down on the keyboard

It may likely be some other obscure mod conflicting with this, but pressing up and down while focusing on the search bar does nothing [in my modpack](https://modrinth.com/modpack/bubble-cobble-redux/version/1.9.3).
EMI `1.1.22+1.21.1+neoforge`

PR #1043, opened 2025-09-16
Add creative only cheat mode (fix #546)

### LINKED ISSUE #546: [Feature Request] Cheat mode: "Creative only"
As a person who is OPped on a survival server, and frequently switches between said server and my creative testing world, it's incredibly annoying to have to switch between cheat modes when I'm using both "cheat to cursor" and "view recipe" on left click. REI, which I've used pretty much exclusively up until now, has a third Cheat mode specifically for that issue, that only allows cheating items in when in Creative mode. I would very much like to switch to EMI because the recipe tree feature is absolutely amazing, but this issue breaks my workflow quite dramatically. This seems like a pretty trivial change to implement, and I'm sure it would improve QoL for quite a number of people.

mageowl:
Converts the cheat mode config option to an enum that allows setting it to "Creative Only", as well as "Always" or "Never". When set to "Creative Only" cheats will only be enabled when you are in creative mode.


Josh65-2201:
This should open recipe view when in survival and cheat mode is set to creative mode

mageowl:
It does.

Josh65-2201:
Doesn't for me, nothing happens when left clicking, either set to cheat one to inventory or cursor.

mageowl:
Did you change the Cheat Mode config to "Creative Only"? If so, what version of Minecraft and what modloader are you using?

Josh65-2201:
Yes, Fabric 0.17.2 on 1.21.1

mageowl:
That's odd. It works on my computer. Did you correctly install the pull request? If so, are there any errors in the console?

Josh65-2201:
Nothing is in console. No idea what you mean by correctly, I cloned your fork, ran `gradlew.bat build` the 1.21 branch, and replace EMI with that builds 1.1.22-snapshot output.

mageowl:
Try using `gradlew.bat neoforge:jar` or `gradlew.bat fabric:jar`. I don't think build will work.

mageowl:
Nevermind. Build seems to work correctly, however "Creative Only" works fine for me.

xplat/src/main/java/dev/emi/emi/config/CheatMode.java
```diff
@@ -0,0 +1,42 @@
+package dev.emi.emi.config;
+
+import dev.emi.emi.EmiPort;
+import net.minecraft.client.MinecraftClient;
+import net.minecraft.text.Text;
+
+public enum CheatMode implements ConfigEnum {
+    ALWAYS("always"),
+    NEVER("never"),
+    CREATIVE_ONLY("creative_only")
```
emilyploszaj: I'd prefer just "CREATIVE" here for both the enum and name, make sure to update lang.

xplat/src/main/java/dev/emi/emi/config/CheatMode.java
```diff
@@ -0,0 +1,42 @@
...
+    public String getName() {
+        return name;
+    }
+
+    @Override
+    public Text getText() {
+        return EmiPort.translatable("emi.cheat_mode." + name.replace("-", "_"));
+    }
+
+    public boolean isEnabled(MinecraftClient client) {
```
emilyploszaj:
```java
return switch (this) {
  case ALWAYS -> true;
  ...
```

xplat/src/main/java/dev/emi/emi/config/EmiConfig.java
```diff
@@ -49,7 +49,7 @@ public class EmiConfig {
 
 	@Comment("Whether cheating in items is enabled.")
 	@ConfigValue("general.cheat-mode")
-	public static boolean cheatMode = false;
+	public static CheatMode cheatMode = CheatMode.CREATIVE_ONLY;
```
emilyploszaj: This will cause existing cheat settings to be overwritten, there's currently no elegant system for having upgrades with changing names, but do you think this is reasonable?

xplat/src/main/java/dev/emi/emi/config/CheatMode.java
```diff
@@ -0,0 +1,42 @@
...
+            }
+            case CREATIVE_ONLY -> {
+                return client.player == null || client.player.isInCreativeMode();
+            }
+            default -> {
+                return false;
+            }
+        }
+    }
+}
```
emilyploszaj: Newline at the end of this file

xplat/src/main/java/dev/emi/emi/config/CheatMode.java
```diff
@@ -0,0 +1,42 @@
...
+
+public enum CheatMode implements ConfigEnum {
+    ALWAYS("always"),
+    NEVER("never"),
+    CREATIVE_ONLY("creative_only")
+    ;
+
+    public final String name;
+
+    CheatMode(String name) {
```
emilyploszaj, line 14: `private` here

xplat/src/main/java/dev/emi/emi/api/EmiApi.java
```diff
@@ -48,7 +48,7 @@ public static EmiRecipeManager getRecipeManager() {
 	}
 
 	public static boolean isCheatMode() {
```
emilyploszaj, line 50: Maybe all calls inside the EMI codebase should use `EmiApi.isCheatMode()` instead of accessing the config itself and passing its own client, thoughts?

emilyploszaj (changes requested):
Mostly just nits, the implementation is reasonable. Two questions, do you think upgrading cheat mode false to creative only by default is sensible? I don't have direct qualms but I'm curious why you chose it instead of the representative option. Secondly, one reason I've put this off is I'm worried about either granting creative cheating perms in modpacks where that was designed to not be the case (even though users can just edit their own config) and making everyone who installs this update required to re-enable cheating. It's been a minute since I've touched the config system, is there a reasonable solution for forwarding old options?

xplat/src/main/java/dev/emi/emi/config/CheatMode.java
mageowl: Should this be changed in `HelpLevel` too?

xplat/src/main/java/dev/emi/emi/api/EmiApi.java
mageowl: I agree, that would make it a lot easier in the future to change this method.

xplat/src/main/java/dev/emi/emi/config/EmiConfig.java
mageowl: If we rename Always and Never to true and false it will be able to interpret booleans correctly.

mageowl:
I think that setting the default to be "creative only" is probably the most sensible option. In vanilla, this is basically the default, where you can only cheat in items from your inventory with creative mode. It feels natural for EMI to follow this rule. As for updating configs, I think the easiest way to fix it would be to rename the always and never variants to true and false. This would make it so that older config values would still be interpreted the same.

mageowl:
Additionally, creative only makes it a lot easier for new players to find cheat mode.

xplat/src/main/java/dev/emi/emi/config/CheatMode.java
emilyploszaj: Just in added code is fine, but I should fix it myself at some point.

xplat/src/main/java/dev/emi/emi/config/EmiConfig.java
mageowl: I went ahead an implemented this fix.

mageowl:
@emilyploszaj I think all of the requested changes have been made. LMK if there is anything else I need to do for this.

PR #1062, opened 2025-10-20
Add a keybind to pull the hovered favorite into the player's inventory

MellowArpeggiation:
Taking notes from GTNH NEI, this PR adds a keybind that - when pressed while hovering over a bookmarked favorite - will automatically search the currently open inventory (chest, crate, barrel, etc) and pull that item into the players inventory, ready for crafting. Works well with synthetic favorites that have a count, which will pull exactly that amount, if available.

Leaving as a draft for now to gauge interest, as I will have to create a new registry for mods to handle their own special inventory containers (like AE2, that doesn't have real container slots, and will need compat added). I've created a quick mock of a new registry method and am testing AE2 side integration with it, would look like the below:

```java
public interface EmiRegistry {

...

	/**
	 * Adds an EmiStackPuller to screens of a given class.
	 * Stack pullers take items from the currently open container and place them in the players inventory.
	 */
	<T extends Screen> void addStackPuller(Class<T> clazz, EmiStackPuller<T> puller);

...

}
```

```java
public interface EmiStackPuller<T extends Screen> {

	/**
	 * Pull items onto the player cursor
	 */
	public void pullStackFromSlot(T screen, int slotId);

}
```

Video demonstration in Monifactory - Forge 1.20.1

https://github.com/user-attachments/assets/4de3b104-8097-4593-8c97-a19d5ee37132

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/api/EmiApi.java
+++ b/xplat/src/main/java/dev/emi/emi/api/EmiApi.java
@@ -123,7 +123,7 @@ public class EmiApi {
 		EmiRecipeManager manager = EmiApi.getRecipeManager();
 		setPages(manager.getCategories().stream().collect(Collectors.toMap(c -> c, c -> manager.getRecipes(c))), EmiStack.EMPTY);
 	}
-
+	
 	public static void displayRecipeCategory(EmiRecipeCategory category) {
 		setPages(Map.of(category, getRecipeManager().getRecipes(category)), EmiStack.EMPTY);
 	}
--- a/xplat/src/main/java/dev/emi/emi/api/EmiApi.java
+++ b/xplat/src/main/java/dev/emi/emi/api/EmiApi.java
@@ -25,6 +25,7 @@ import dev.emi.emi.recipe.EmiSyntheticIngredientRecipe;
 import dev.emi.emi.recipe.EmiTagRecipe;
 import dev.emi.emi.registry.EmiRecipes;
 import dev.emi.emi.registry.EmiStackList;
+import dev.emi.emi.registry.EmiStackPullers;
 import dev.emi.emi.runtime.EmiFavorite;
 import dev.emi.emi.runtime.EmiHistory;
 import dev.emi.emi.runtime.EmiSidebars;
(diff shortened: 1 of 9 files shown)


MellowArpeggiation:
I got it working with AE2, changes are AE2 side (new packet, etc) so I'll be making a PR there too, will link here when done

MellowArpeggiation:
And donezo! Check out the accompanying PR at https://github.com/AppliedEnergistics/Applied-Energistics-2/pull/8689

emilyploszaj:
Please reduce the diff noise and reformatting in edited classes. The proposed API is far too targeted, please design a generic solution that uses API concepts like `EmiStack` and `EmiIngredient`, does not directly interface with slots when not necessary, and has further context. The existing recipe handler should be a good example. Additionally, why should this be a separate class from recipe handlers? They seem to overlap significantly. Could this not be an alternative "craft" mode?

MellowArpeggiation:
Thanks for the review, I can revert the whitespace changes, I had a post-proc removing empty whitespace that I left on (globally) from a previous project, and thought at the time that it was both minor and a good spot of cleanup.

The reason the API interfaces with slots directly is it is the simplest solution that can work with almost any inventory, regardless of how it is implemented (it is also almost identical to the existing auto-crafting item moving implementation!), the only case it fails in is with fake slots, like AE2 slots, which is what the registry is for.

This PR is not a recipe handling solution, I gave it the same separation the `EmiStackProvider` does, since it is generally only to be used to patch up inventories that don't allow automatic interaction, not to provide recipe support. Note in `EmiRegistry` it even follows the same implementation pattern as the generic stack provider!

That said, I'll make those changes to the API itself to remove the `ItemStack` reference and use the EMI types where possible, and provide some more useful context for item pull providers.

Again, thanks for reviewing, this functionality is something I _sorely_ miss from the currently maintained 1.7.10 version of NEI, and I am happy to make any changes necessary to get it approved!

MellowArpeggiation:
On second review, maybe the API registration could be rolled _into_ `EmiStackProvider`, I just don't want to break existing implementers if possible, what do you think?

MellowArpeggiation:
Okay I'm ready for a second review, I have:
* Removed all unnecessary whitespace changes
* Switched the API registry to use `EmiStack` (and tested with AE2 integration, both 1.20 & 1.21)
* Removed instances of `ItemStack` in `EmiApi`, replacing with `EmiStack`. preserving behaviour

I investigated the possibility of adding more context to the API registration interface, but all the relevant context (like player inventories and the like) are all provided already via more standard means. If you can think of any appropriate context to be added via arguments please let me know and I can add them.

Oh and I'm still unsure about the use of `if (condition) continue;` as I brought up in the discord, if you want that to be a full block instead of inline I can change that too.

Issue #263, opened 2023-07-30
[1.20.1] Suggestion: Add a shortcut function to enable cheat mode while holding CTRL and clicking the EMI config button.


Gbergz:
## Suggestion: 
Add a shortcut function to enable cheat mode while holding CTRL and (left) clicking the EMI config button. 
(Only possible with OP perms, ofcourse, maybe a message in chat if not op as a reminder.)

A quality of life improvement. Mainly for modpack developers or whoever needs it.

![pic](https://github.com/emilyploszaj/emi/assets/26634554/8d9ab7ea-f0f7-4f38-b02f-7737de7311ac)

**-** Gbergz

PR #1091, opened 2025-12-09
Add switching cheat mode when clicking EMI button with ctrl

Taskeren:
Like JEI, when click the EMI button with ctrl pressed, it toggles the cheat mode.
I don't if I need to do something to trigger config save or something after changing the value.

And I'm sorry that my code formatter is much different to the existing style, so I'll leave it unformatted.

#263

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/screen/EmiScreenManager.java
+++ b/xplat/src/main/java/dev/emi/emi/screen/EmiScreenManager.java
@@ -7,6 +7,7 @@ import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import dev.emi.emi.config.CheatMode;
 import net.minecraft.command.argument.ItemStackArgument;
 import net.minecraft.component.ComponentChanges;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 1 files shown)


Taskeren:
I also want to make the button looks different depending on whether cheat mode is on or not.
But couldn't find how EMI render the text. Much appreciate if somebody can help!

emilyploszaj:
Please do best effort to match code style in the repo and use existing utilities for keys instead of vanilla options, additionally, this doesn't handle all of the config values nor explain the behavior. Should this be the only config option that can be quick swapped in EMI by pressing the config button? EMI is not aiming to be JEI, so just because that mod does something is not justification here.

Taskeren:
> Please do best effort to match code style in the repo and use existing utilities for keys instead of vanilla options, additionally, this doesn't handle all of the config values nor explain the behavior. Should this be the only config option that can be quick swapped in EMI by pressing the config button? EMI is not aiming to be JEI, so just because that mod does something is not justification here.

it's a shortcut way to swap cheat mode value, but the option now has more than 2 values, which I don't have a good idea to handle this. So I'd consider CREATIVE as a default value, and once player has used this feature, we only swap it in TRUE and FALSE.

the reason I want it in EMI is for use cases that players like me, who don't want to fully non-cheating in the game, want to grab something out from EMI during the gameplay. In this case, I would stay in survival mode, so the CREATIVE doesn't work, and changing the value in the config screen is much troublesome imo.

Taskeren:
And I'm not sure if I missed it, I don't see a util for checking if Control is down. Can you tell me where it is?

PR #1092, opened 2025-12-09
Add switching cheat mode when clicking EMI button with ctrl (1.20 branch)

Taskeren:
Same as #1091 but for 1.20 branch

Like JEI, when click the EMI button with ctrl pressed, it toggles the cheat mode.
I don't if I need to do something to trigger config save or something after changing the value.

And I'm sorry that my code formatter is much different to the existing style, so I'll leave it unformatted.

#263


emilyploszaj:
There is no need to PR to other versions manually, I'll review the existing PR for latest

Issue #1041, opened 2025-09-10
[Feature Request] Multitree / Add Tree


PlayfulPiano:
Currently, EMI is restricted to calculating and visualizing a single recipe tree at a time. A big limitation is when you want to craft multiple items and determine the shared ingredient demands at once, e.g. building a multiblock in GT, or a schematic for Create.

My request would be to implement the ability to include multiple different items to the same recipe tree, by allowing the user to shift + click on the view recipe tree button for a recipe, which would then give a + icon for adding that to the current recipe tree.

While in the recipe tree UI screen, the root item will now be a row of items you have included, where you can click on any of the items to view/swap to their corresponding recipe tree, or use your scroll wheel to scroll through the root items currently selected. This is to avoid the likely insane coding requirement to implement a DAG styled tree. Next to each side of the root item row would be arrows for manual scrolling. Mousing over a root item would also have a red rectangle appear above it with an X, for deleting that root item tree.

As for synthetic favorites, those quantities would be based on matching the root item demands. The synthetic favorite quantities would be summed together across root item trees. As for instances where one root is an ingredient for another root, synthetic favorites would calculate based on making the full quantity of each root independent of each other.

E.g. making 64 iron ingots and 15 iron buckets would have a synthetic favorite for getting 64 + 45 = 109 iron ore.

Thank you for reading.

lucasteo16:
I was literally looking for this, this is useful especially for cases where a lot of common material is used. 

E.g. Crafting a variety of items that shares some common intermediate materials. The normal workflow for a single tree would be gathering specific amount of those materials, craft those materials just for the first tree once that's done, you proceed to the next tree just to realize that the next item needs the exactly the same intermediate materials again. 

In contrast, this will allow building out all the trees at once, gather all stuff at once and then craft a bunch of different things at once.
No going backand forth repeatedly for the same ingredient required for different trees.

I'm wondering how the potential implementation could be, I'm thinking something like letting the virtual bookmarks to sum all the require intermediate material together when more than one trees are in the tree list. On the recipe page, clicking tree button will now add a new tree to the list. Removal of a tree can simply be done by removing it from the bookmark. The tree editing page should allow navigation to other trees (so each tree editor will only show one tree and you jump between other trees with some new ui, instead of showing more than one trees in a single editor which I speculate that would require more changes on existing iimplementati, the tree editor-thingy can just be reused for each tree in a selectable list of trees)

gabbieblue:
[...]

PR #1100, opened 2025-12-26
Add multiple recipe tree management and batch manipulation

ICaxapI:
Enhance recipe tree functionality with support for multiple trees, allowing addition, removal, and selection. Add combinable costs and progression for aggregated handling. Introduce new tooltips for advanced batch adjustments with `Ctrl` and `Alt`.

To add new recipe trees, you need to press the add recipe button in the new recipe while holding down the Shift key.
image


emilyploszaj:
Sorry to point you to a discussion forum like Discord, but there has been notable work on features overlapping these behaviors there which would be a good indicator to inform how design should look. I do not have time to thoroughly review this code at the moment, but I will say an implementation already exists that takes many different design decisions. Could you justify any changes you've made to the user experience (visuals, keybinds, tooltips, positioning, functionality) in your implementation?

ICaxapI:
Hi! I read CONTRIBUTING.md, but I couldn't find a Discord link for syncing the functionality/disign.
This [FR](https://github.com/emilyploszaj/emi/issues/1041) is largely a response to a request for multi-tree functionality. You can read the original request there, and this PR mostly fulfills that request.
This PR only adds functionality for displaying and counting materials for multiple items at once, accounting for cross-materials and recipes. Here, I've only added a bind for pressing the tree display button with shift, which will add the selected item to the current tree in memory and display two trees at once. (This picture from another PR, there I slightly improved the visuals and now it looks like this. If my PR is in demand, I could update it so that it looks correct and more polished, just give me a sign :) ) ->
image

In another PR, I also added saving/renaming/deleting tree sets to selectable keybinds, which will allow me to store multiple recipe sets at once and quickly switch between them (in a separate page or subpanel). I reused one class from that PR for display, so it was in my PR, but I can transfer all this extended functionality for [multi-trees only] to this PR. I didn't want to do this because I would have given credit to the author of the bookmark PR for the class I reused.
image
image

If there are any specific areas that need improvement (for example, in the design), please describe them. I would like to see this functionality in the official EMI.

ICaxapI:
After three weeks of testing, I can say it works perfectly. I've already used this patch to plan and craft huge factories with multi-block machines and complex, multi-tiered recipes. Synthetic favorite lists reached halfway down the second page, and everything worked perfectly, including saving the factory plan between sessions.

xplat/src/main/resources/assets/emi/lang/fr_fr.json
```diff
@@ -316,7 +316,9 @@
   "tooltip.emi.synfav.partially_craftable": "§dImpossible§r de tout faire\n§6Cliquez§r pour en fabriquer",
   "tooltip.emi.synfav.fully_craftable": "§aPossible§r de tout faire §a(%1$s)§r\n§6Cliquez§r pour en fabriquer",
 
-  "tooltip.emi.bom.batch_size": "Quantité: %s\nScrollez pour changer\nMaintenez §6[shift]§r pour changer par 16",
+  "tooltip.emi.bom.batch_size": "Quantité: %s\nScrollez pour changer\nMaintenez §6[Shift]§r pour changer par 16",
+  "tooltip.emi.bom.batch_size.multiply": "Maintenez §6[Ctrl]§r pour ajouter un multiplicateur",
+  "tooltip.emi.bom.batch_size.all": "Maintenez §6[Alt]§r pour appliquer à tous les objets racine",
```
Starmania: This is not fluent french ("racine" for "root" does match, but feels a bit off)

Starmania: Here are some better alternatives imo
- objets principaux
- objets finaux

Zeyrox902:
I’ve identified a few implementation errors:
- Leftover handling: If an item is marked as a leftover for one recipe, it isn't being reused for others.
- Root items: Similarly, a leftover item cannot currently be used as a root item.

UI Suggestion: When the total cost is very high, the interface zooms out significantly to fit everything on screen, making it hard to read. Implementing a "break" for large scales would be a great improvement, though it might be outside the scope of this PR.

Screenshots for context, illustrating the second point:

image
image

ICaxapI:
> I’ve identified a few implementation errors:
> 
> * Leftover handling: If an item is marked as a leftover for one recipe, it isn't being reused for others.

Yes, it's true leftovers don't count as ingredients for crafting adjacent trees. I actually considered this during the implementation and thought it would be more convenient. But now I realize I haven't encountered a single issue in the game where this was truly useful and didn't become annoying. It's actually a very easy fix.

> * Root items: Similarly, a leftover item cannot currently be used as a root item.

If you mean that if one root recipe contains a final element from another root recipe, it won't be calculated correctly, then this is the same issue as above. I'll fix it in the next commit. Thank you for your attention.

ICaxapI:
> UI Suggestion: When the total cost is very high, the interface zooms out significantly to fit everything on screen, making it hard to read. Implementing a "break" for large scales would be a great improvement, though it might be outside the scope of this PR.

Yep, with multitrees this line can get really long. Maybe we should wrap elements to the next line when the number of elements exceeds a certain limit?

PR #1102, opened 2025-12-29
Split render methods

ICaxapI:
Split render methods into `renderBase` and `renderAmount` to fix the wrong order in rendering.

Now, in batching renderer enabled and with large trees numbers may render under icons, witch cause the visual bug ->
image

This PR fix this bug ->
image

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/screen/BoMScreen.java
+++ b/xplat/src/main/java/dev/emi/emi/screen/BoMScreen.java
@@ -227,10 +227,10 @@ public class BoMScreen extends Screen {
 				context.drawCenteredText(EmiPort.translatable("emi.leftovers"), 0, cy - 16 + 40);
 			}
 			for (Cost cost : costs) {
-				cost.render(context);
+				cost.renderBase(context);
 			}
 			for (Node node : nodes) {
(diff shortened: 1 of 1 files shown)


emilyploszaj:
This is a bit of an annoying workaround of a problem introduced by other mods. Shown behavior is not present in vanilla. Should this really be in EMI and not resolved elsewhere?

ICaxapI:
> Это несколько досадное обходное решение проблемы, возникшей из-за других модов. Показанное поведение отсутствует в оригинальной игре. Действительно ли это должно быть реализовано в EMI, а не в других модах?

I could be wrong, but I think the error isn't caused by other mods, but by large trees and internal rendering synchronization issues. Vanilla Minecraft doesn't have crafting trees large enough to reproduce this issue. I just tested this on a pure Minecraft + gregtech + emi setup, and it broke. I also reproduced it on a pure Minecraft + my mod + emi setup, and it caused the issue again. The error only occurs on fairly large trees, but not every time, so I think it's a synchronization issue.

This seems to be a fairly simple fix that follows the original logic without complicating it, while also fixing rendering issues with the EMI recipe tree screen. Honestly, I can't even imagine how this could be fixed using other mods' code.

PR #1134, opened 2026-03-05
i18n: Traditional Chinese localization.

CrazyO9:
Dear Manager,
I am currently working on localizing the Traditional Chinese language pack for TFG (TerraFirmaGreg) and would like to contribute to this project.
Regarding the current progress, I have a specific technical question:

* Version Comparison: I have not yet confirmed whether there are differences in the translation strings between versions v1.21 and v1.20.5.
* Compatibility Goal: If possible, I would like to ensure the text is compatible with both versions to maintain consistency for the community.

I am eager to apply the translated text and would appreciate your guidance on the best way to proceed.


emilyploszaj:
The latest version will always have the correct translations, and I will handle merging everything backwards.

emilyploszaj (approved):

Issue #1040, opened 2025-09-10
[Mod Compat] Occultism and Ars Nouveau recipe tree issues


Mojken:
The Dimensional Storage Actuator and Storage Accessor from Occultism, as well as the Storage Lectern from Ars Nouveau, seem to not interact well with EMI's recipe tree. The tree doesn't detect items in the open inventory, and doesn't craft items when they are clicked.

The Storage Lectern seems to be able to pull *some* items - I was testing with a comparator and it put the quartz in, but nothing else, even when I had the rest in my own inventory.

P.S: As for the Occultism stuff, it would be very cool if it could also handle other machines attached via Djinni. What machine is attached is available in the interface, but would require specific compatibility to properly detect and use. What I mean is that beyond just the crafting table support, it could also issue stone to be smelted if you click that step of the tree.

EDIT: ~~Perhaps I opened this too hastily, there's a *closed* issue regarding this [over at occultism](https://github.com/klikli-dev/occultism/issues/1255).~~ Turns out it's not specifically about the recipe tree. The plus button does work as expected, thinking about it it's rather surprising that this is an issue at all given that.

PR #1149, opened 2026-04-02
[1.19.2] Fix JEMI subtype handling for JEI 11.7+

### LINKED ISSUE #1146: Searching Items/fluids with NBT return recipes for the base item, not the specific nbt variant.
### Describe the bug, attach relevant log files, and describe what environment the issue occurs in

We've been noticing some incompatibilities with JEI after versions 11.7 (forge 19.2) that apparently need to be fixed in EMI.

Full details on the issue are here: https://github.com/EnigmaticaModpacks/Enigmatica9/issues/1040

Apparently someone has published a patch for this on github here: https://github.com/StarskyXIII/EMI_Patches/releases/tag/1.0

I've asked @StarskyXIII to consider PRing it to you for a fix in EMI directly.

StarskyXIII:
fix #1146

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -16,4 +16,4 @@ fabric_api_version=0.76.0+1.19.2
 
 forge_version=1.19.2-43.2.8
 
-jei_version=jei-1.19.2-fabric:11.6.0.1015
+jei_version=jei-1.19.2-fabric:11.7.0.1025
--- a/xplat/src/main/java/dev/emi/emi/jemi/JemiPlugin.java
+++ b/xplat/src/main/java/dev/emi/emi/jemi/JemiPlugin.java
@@ -50,7 +50,6 @@ import mezz.jei.api.constants.VanillaTypes;
 import mezz.jei.api.ingredients.IIngredientType;
 import mezz.jei.api.ingredients.IIngredientTypeWithSubtypes;
 import mezz.jei.api.ingredients.ITypedIngredient;
-import mezz.jei.api.ingredients.subtypes.IIngredientSubtypeInterpreter;
 import mezz.jei.api.ingredients.subtypes.ISubtypeManager;
 import mezz.jei.api.ingredients.subtypes.UidContext;
 import mezz.jei.api.recipe.RecipeIngredientRole;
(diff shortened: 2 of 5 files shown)


emilyploszaj:
Why did you remove the more invasive check? Additionally, is this compatible with modern versions? If so, please only target the latest version branch.

StarskyXIII:
I removed that invasive check because it was only really needed for older JEI internals.

On 11.6, EMI had to poke into `SubtypeRegistration` / `SubtypeInterpreters` to answer “does this ingredient type have subtype handling?”. In 11.7+, JEI exposes that directly via `ISubtypeManager#hasSubtypes(...)`, and the internal path changed, so using the public API seemed safer and less brittle.

As for modern versions, this looks like it’s already handled on the latest branch, so my intent here was just to backport the fix to 1.19.2.

emilyploszaj:
I see, perhaps it'd still be better to leave the old path to just silently fail? That way EMI could remain compatible with older JEI versions in 1.19.2

StarskyXIII:
That’s fair.

From my perspective, keeping compatibility with older JEI versions on 1.19.2 does not seem especially meaningful here.

As mentioned in Enigmatica 9 [#1040](https://github.com/EnigmaticaModpacks/Enigmatica9/issues/1040), the last working version was JEI 11.6.0.1024, but downgrading to that would also require downgrading Almost Unified to 0.8.1, which was marked as archived on CurseForge.

Because of that, I don’t really see a practical use case for preserving support for old JEI in this branch. If a modpack is willing to update EMI to a newer build, it seems unlikely they would intentionally keep JEI on an older version at the same time.

That said, this is your mod, so I’m happy to follow whichever direction you think is best.

emilyploszaj:
There's almost no cost to it and it's not maintained on new versions, yeah, please keep it around.

StarskyXIII:
Sorry for the delay. I've been tied up with other things these past few days

emilyploszaj (approved):

emilyploszaj:
Don't worry at all about your delay, thanks for the contribution.

PR #1161, opened 2026-04-28
Refactor workstation handling in RecipeScreen for overflow management

### LINKED ISSUE #948: Is there any way to increase the number of workstations shown in EMI
Some mods like GregTech have more than 10 workstations for one kind of recipe, but EMI can only show a limited number of workstations. For examples there are 11 workstations for Forge Hammer recipe in GregTech, but EMI only shows 8 workstations.
in JEI:

Image

in EMI:

Image

So is there any way to show all the workstations for one kind of recipe like JEI or at least increase the number of workstations shown in EMI?

### LINKED ISSUE #999: [Bug] EMI does not showcase all matching workstations if it goes beyond 10.
When looking at a recipe in EMI, you can only see the first 10 workstations that match that recipe, with others not being visible whatsoever (even though right clicking on those workstations will still show the recipe group). Biggest case example would be gregtech machines, due to them usually having over 10 single/multiblock machines for the same recipe process.

Here's a comparison between EMI (Star Technology w/TMRV) and 1.7.10 NEI (GTNH):

Image

Image

link-fgfgui:
Refactored workstation loop to handle overflow correctly.

Closes #999, Closes #948.

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/screen/RecipeScreen.java
+++ b/xplat/src/main/java/dev/emi/emi/screen/RecipeScreen.java
@@ -383,9 +383,15 @@ public class RecipeScreen extends Screen {
 			List<EmiIngredient> workstations = EmiApi.getRecipeManager().getWorkstations(tab.category);
 			if (!workstations.isEmpty()) {
 				WidgetGroup widgets = new WidgetGroup(null, 0, 0, 0, 0);
-				for (int i = 0; i < workstations.size() && i < getMaxWorkstations(); i++) {
+				int maxWorkstations = getMaxWorkstations();
+				for (int i = 0; i < workstations.size() && i < maxWorkstations; i++) {
 					Bounds bounds = getWorkstationBounds(i);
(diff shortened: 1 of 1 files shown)


link-fgfgui:
<details>
  <summary>Screenshots</summary>


</details>

emilyploszaj (approved):
Looks good, thanks

Taskeren:
Thanks for the change.

However, I want to propose that when the last slot is clicked, the whole workstation list is expanded instead of only the truncated ones.
And if possible, give the expanded list a name like `Workstation for {Cateogry}` like **Workstation for Item to Chemical**.

And @emilyploszaj can you please port this PR back to 1.20.1 and release the update (also the 1.21.1)?

link-fgfgui:
Instead of replying to the closed PR, it might be better to open a new issue so we can track this properly.
About Porting, here is the answer:
https://github.com/emilyploszaj/emi/blob/35392222b42e82655c06c0a5bd8897c40e445a23/CONTRIBUTING.md?plain=1#L17

PR #1194, opened 2026-05-22
fix: resolve integer overflow on maxAmount (#1191)

### LINKED ISSUE #1191: Shift + Fill Recipe (Shift + '+') clears vanilla crafting grid on ambiguous recipes
### Minecraft Version

1.21.1

### Mod Loader

NeoForge 21.1.229

### EMI Version

EMI 1.1.24+1.21.1+neoforge

### Description

When using EMI's recipe transfer feature with `Shift + +` ("fill max") on recipes that have multiple possible outputs from the same input pattern, the vanilla crafting table gets cleared instead of being filled.

The issue only happens with `Shift + +`.

Normal `+` works correctly.

JEI does not have this issue under the same conditions.

### Reproduction Steps

1. Install:

   * EMI
   * Quark
   * Farmer's Delight

2. Open a vanilla crafting table.

3. Put 9 potatoes in a 3x3 pattern.

4. Hover the resulting crate item and press `R`.

5. In EMI recipe view:

   * click normal `+`
     -> works correctly
   * click `Shift + +`
     -> crafting grid is cleared and items return to inventory

### Important Details

Seems, this only happens on recipes where:
* multiple mods provide the same crafting pattern
* but produce different outputs

Example:
* Farmer's Delight potato crate
* Quark potato crate

Both use:
* 3x3 potatoes

### Additional Observations

The issue does NOT happen in the Sophisticated Backpacks crafting upgrade.

Using the exact same recipe and the same `Shift + +` behavior:
- vanilla crafting table -> broken
- Sophisticated Backpacks crafting grid -> works correctly

This suggests the issue may be specific to EMI's handling of the vanilla crafting container / recipe transfer logic.

### Additional Testing

Tested configurations:

* EMI only -> issue present
* EMI + EMI Loot -> same issue
* JEI only -> works correctly

### Expected Behavior

`Shift + +` should fill the crafting grid to the maximum possible amount, just like normal `+`, without clearing the existing crafting grid.

### Actual Behavior

The crafting grid contents are returned to the inventory and the crafting grid becomes empty.

### Notes

The issue seems specific to EMI's "fill max" behavior and not to normal recipe transfer.

link-fgfgui:
fix #1191

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/registry/EmiRecipeFiller.java
+++ b/xplat/src/main/java/dev/emi/emi/registry/EmiRecipeFiller.java
@@ -212,8 +212,11 @@ public class EmiRecipeFiller {
 						maxAmount = Math.min(maxAmount, ui.max);
 					}
 				}
-				maxAmount = Math.min(maxAmount, amount + batchesAlreadyPresent(recipe, handler, screen));
-
+				int batches = batchesAlreadyPresent(recipe, handler, screen);
+				if (amount < Integer.MAX_VALUE - batches) {
+					maxAmount = Math.min(maxAmount, amount + batches);
+				}
+				
 				if (maxAmount == 0) {
 					return null;
 				}


emilyploszaj:
Sorry I missed this for so long, what is causing the underlying integer overflow here?

link-fgfgui:
The underlying integer overflow is caused by the "fill max" (`Shift + +`) feature, which explicitly passes `Integer.MAX_VALUE` as the `amount` of batches to fill.

Here is the exact breakdown of how the overflow happens based on the related code:

1. **Passing `Integer.MAX_VALUE`:**
   When a user Shift-clicks the fill button or uses the corresponding keybind, the `amount` passed is `Integer.MAX_VALUE`. 
https://github.com/emilyploszaj/emi/blob/5b49fb13228088469d32fc8cfbbd6190e7cfc0be/xplat/src/main/java/dev/emi/emi/api/widget/RecipeFillButtonWidget.java#L72

2. **The Overflow in `EmiRecipeFiller.java`:**
   This `amount` is passed down the chain (e.g. through `StandardRecipeHandler.java` via `context.getAmount()`) and reaches the calculation in `EmiRecipeFiller.java`. Before this PR, `maxAmount` was calculated directly:
https://github.com/emilyploszaj/emi/blob/5b49fb13228088469d32fc8cfbbd6190e7cfc0be/xplat/src/main/java/dev/emi/emi/registry/EmiRecipeFiller.java#L215
   If the vanilla crafting grid already has some items matching the recipe pattern, `batchesAlreadyPresent(...)` returns a positive number (e.g. `> 0`). Because `amount` is exactly `Integer.MAX_VALUE`, evaluating `amount + batches` causes an **integer overflow** and wraps around to a negative number (e.g., `Integer.MIN_VALUE + (batches - 1)`).


3. **The Resulting Bug:**
   The actual filling logic, found in `EmiRecipeFiller.clientFill(...)`, is designed to work in two steps. First, it clears the existing crafting grid using a `QUICK_MOVE` action (returning existing items to the inventory):
   ```java
   List<Slot> clear = handler.getCraftingSlots(screenHandler);
   for (Slot slot : clear) { ... manager.clickSlot(..., SlotActionType.QUICK_MOVE, player); }
   ```
   Second, it iterates through the expected stacks to fill the slots. However, the number of items needed is taken from the negatively-counted stack:
   ```java
   int needed = stack.getCount(); // This is now a negative number!
   ```
   Since `needed` is negative, the filling operations (like `while (needed > 0)`) are completely skipped. The game clears the grid but places absolutely nothing back in. This perfectly explains the exact buggy behavior reported in #1191.

This PR fixes it by adding a safe bounds check (`amount < Integer.MAX_VALUE - batches`) before performing the addition. If `amount` is already at max capacity, it skips the addition so `maxAmount` stays correct without wrapping around into negative values.

PR #1214, opened 2026-06-25
[1.21.1] Add compatibility with nbt/component driven recipes

0196071:
Add compatibility with nbt/component driven recipes, such has kubejs nbt driven recipe components.
Currently only has us localization.

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/config/EmiConfig.java
+++ b/xplat/src/main/java/dev/emi/emi/config/EmiConfig.java
@@ -78,6 +78,10 @@ public class EmiConfig {
 	@ConfigValue("general.search-tags-by-default")
 	public static boolean searchTagsByDefault = false;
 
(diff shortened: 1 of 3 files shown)


emilyploszaj:
Sorry this is far too blunt of a solution to accept, requiring component matches for every lookup will break more recipes than it'll fix. If you'd still like to try your hand resolving this issue, adding a data-driven method for setting comparisons would be interesting.

0196071:
I see your point. The issue I ran into was from a friend's recipe chain that technically uses the same item id for every step, but stores the current step as NBT/component data.

When you mention adding a data-driven method for setting comparisons, do you mean something like a datapack JSON or item tag that would let pack makers mark specific items as needing component comparison?

So EMI would only use Comparison.compareComponents() for those specific items, instead of changing the default behavior globally.

Issue #1210, opened 2026-06-19
[1.1.24+1.20.1] "Critical error occured during reload" in hevily modded enviroment


yuusyaasisutanto:
I tried so hard but can't identified what mod causing this...
I think "java.lang.NullPointerException: Cannot invoke "dev.emi.emi.api.recipe.EmiRecipeCategory.getId()" because "category" is null" has cause this issue.

WARNING: too heavy file (10MB)
[latest.log](https://github.com/user-attachments/files/29143122/latest.log)

yuusyaasisutanto:
Identified: Mekanism Extras 1.5.0
I've downgraded to 1.4.6, and works it so well.

but minimam configuration didn't reproduce this issue, perhaps this issue has caused by mods interaction

liquidcatmofu:
Additional information: I was able to identify one specific source of this error.

In my case, the issue is caused by an incompatibility between:

* Just Enough Mekanism Multiblocks 4.18
* Mekanism Extras 1.5.0

Before EMI fails to reload, Just Enough Mekanism Multiblocks throws the following error while registering its Mekanism Extras integration:

```text
java.lang.NoClassDefFoundError: com/jerry/mekanism_extras/common/registry/ExtraBlock
	at giselle.jei_mekanism_multiblocks.client.jei.category.extras.ExtraMatrixCategory.<init>(ExtraMatrixCategory.java:27) ~[JustEnoughMekanismMultiblocks-1.20.1-4.18.jar%23984!/:4.18]
```

To identify the offending recipes, I locally patched EMI to skip and log recipes whose category is null instead of aborting the entire reload.

This identified eight affected recipes, all created by:

```text
giselle.jei_mekanism_multiblocks.client.emi.EMIMultiblockRecipe
```

Both the recipe ID and category were null:

```text
[EMI] Skipping recipe with null category: id=null, recipeClass=giselle.jei_mekanism_multiblocks.client.emi.EMIMultiblockRecipe
```

After skipping those recipes, EMI successfully completed the reload and baked the remaining recipes.

Downgrading Mekanism Extras from 1.5.0 to 1.4.6 also prevents the issue.

I have reported the incompatibility to Just Enough Mekanism Multiblocks here:

https://github.com/gisellevonbingen-Minecraft/JustEnoughMekanismMultiblocks/issues/38

yuusyaasisutanto:
Many thanks for that!!!

yuusyaasisutanto:
[...]

PR #1222, opened 2026-07-20
Prevent null recipe categories from aborting reload

liquidcatmofu:
### Description

Third-party recipe implementations may return a null category, as seen in #1210. Currently, this causes EMI itself to throw a `NullPointerException` while reporting the unregistered category, aborting the entire recipe reload and making EMI unusable.

This change filters recipes with null categories before they are passed to the recipe manager. It also logs the recipe ID and implementation class, allowing the faulty integration to be identified without preventing the remaining recipes from loading.

This does not fix the underlying third-party integration issue, but prevents a single invalid recipe from aborting the entire EMI reload and provides enough information to diagnose its source.

### Testing

- Built the project successfully.
- Confirmed startup on Minecraft 1.21.1 with Fabric.
- Confirmed startup on Minecraft 1.21.1 with NeoForge.
- Tested the equivalent change against the failure reported in #1210 on MC 1.20.1 Forge.

  - 8 recipes with null categories were identified and filtered.
  - The offending recipe implementation was logged.
  - EMI completed recipe and search baking instead of aborting with a `NullPointerException`.

### DIFF
--- a/xplat/src/main/java/dev/emi/emi/registry/EmiRecipes.java
+++ b/xplat/src/main/java/dev/emi/emi/registry/EmiRecipes.java
@@ -88,6 +88,15 @@ public class EmiRecipes {
 		invalidators.addAll(EmiData.recipeFilters);
 
 		invalidators.add(r -> {
+			EmiRecipeCategory category = r.getCategory();
+			if (category == null) {
+				EmiReloadLog.warn("Recipe " + r.getId() + " from " + r.getClass().getName() + " has null category");
+				return true;
+			}
+			return false;
+		});
+
+		invalidators.add(r -> {
 			for (EmiIngredient i : Iterables.concat(r.getInputs(), r.getOutputs(), r.getCatalysts())) {
 				if (EmiHidden.isDisabled(i)) {
 					return true;


emilyploszaj:
This is probably best handled later in the file during the initialization of the recipe manager https://github.com/emilyploszaj/emi/pull/1222/changes#diff-3be079dd3145a0d1a332fc1f6ca0ba8afec23f0bd2e81f512920cdeaeca36358R173

Additionally, excuse me if there is no need, but if you've used any LLM/generative AI assistance for the code or description here please make sure to disclose that.

Thanks, happy to merge if you move the filtering.

liquidcatmofu:
When testing #1210, as you mentioned, I had placed the code immediately before the section where the `NullPointerException` occurred.

```java
for (EmiRecipe recipe : recipes) {
	Identifier id = recipe.getId();
	EmiRecipeCategory category = recipe.getCategory();
	if (category == null) {
		EmiReloadLog.warn("Skipping recipe with null category: id=" + recipe.getId() + ", recipeClass=" + recipe.getClass().getName());
		continue;
	}
	if (!categories.contains(category)) {
		EmiReloadLog.warn("Recipe " + id + " loaded with unregistered category: " + category.getId());	// NullPointerException
	}
```

However, with this implementation, invalid recipes would remain in `recipes`. Therefore, I believe it is more appropriate to prevent this by filtering the recipes before passing them to the manager, which is why the current implementation is set up this way.
If you still want me to move it, I’ll do so.

Regarding the use of an LLM, I used it to investigate the cause, organize the explanation, and for translation. I apologize for using it without explicitly stating so. I did not use an LLM for this sentence; instead, I used DeepL.

PR #1238, opened 2026-08-12
fix: incorrect paths for maven dependecies

mikesk8r:
Does exactly what the title says. The dependency paths in the README were slightly inaccurate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,22 +23,22 @@ Here are common dependency setups for different loaders and build systems.
 ```gradle
 dependencies {
 	// Fabric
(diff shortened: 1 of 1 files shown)


emilyploszaj:
Appreciate the attention to detail, but this is intentional formatting, EMI's versions strings include general Minecraft version released for, `1.0.2` is not enough to point to a single release, you'd need to do `1.0.2+1.19.2`. However, you can, for instance, use the same EMI jar on 1.21.0 and 1.21.1, so using the consumer's `minecraft_version` could easily solve problems. Thanks for the concern though!

PR #1241, opened 2026-08-15
icon tweaks

Jamiscus:
chagned some icons to be more consistent with vanilla (global settings, dev bug icon, crafting type icons), arrows to match the built in neoforge creative inventory buttons (so it looks consistent on forge), and some (favourites star, microfont Q and G) just because i thought they looked nicer

if you like some of these changes but not others lmk and i can remove some :)

if this is a bad pull request feel free to pelt me with tomatoes and rake me over the coals


emilyploszaj:
Hi, of these changes only the microfont adjustments and the global config icon are something I'd be fine using, if you could revert everything else that'd be fine.

